PR #6, opened 2018-05-05
Re-fix greenhouses crop amounts

inkognito1100:
Fixed not only crop amounts but also resources i/o rates and build-in storages too. (x1.7 for 2.5m and x2.55 for 3.75). 
crop_size = #$../mass$
crop_size *= 1000
doesn't work for me


MoreRobustThanYou:
Did you test it?

MoreRobustThanYou:
Nvm I tested it, it works. Thanks!

PR #7, opened 2018-05-05
Updated profiler GUI to use the Canvas system

PiezPiedPy:
Included the Profiler from Trajectories that was originaly from Kerbalism, I gave it a major overhaul in Trajectories with the following changes:

* The Profiler GUI has been updated to use KSP's PopupDialog class which in turn uses a GameObject with a canvas and also has the added benefit of stopping most mouse click through events, even to the context menus.
* A button to reset the average values.
* A button to disable clearing of values when there are zero calls.
* A frame limiter to smooth out display.
* An average calls per physics frame value and physics frame counter.
* Profiler Addon no longer loads in the release build.

You can change the display framerate by changing the `update_fps` value on line 59.
To Show/Hide the profiler use CTRL+P

There is more info in the CONTRIBUTING.md file on how to use the profiler.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
  * Profile importing is now available to modders who wish to import their own procceses, rules, supplies etc (thanks to PiezPiedPy)
    see Issue #2 on GitHub for more information [Here](https://github.com/MoreRobustThanYou/Kerbalism/issues/2)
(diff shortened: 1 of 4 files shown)


MoreRobustThanYou:
Picture?

PiezPiedPy:
screenshot4

PiezPiedPy:
screenshot3

No reason for this one 😎

PR #10, opened 2018-05-10
Rebalance food

madman2003:
This is an attempt at getting food production back to realistic levels, as well some minor related things that I found along the way. Food production is a huge constraint on any mission beyond low orbit, and the current levels didn't at all match up with reality, especially the buff of SSPX modules was over the top, considering they have habitats inside as well.

It also adds Carbon Dioxide processing, which is what plants do for us.

I'm using the pull request to get a build (so not tested until I say so :-P).

Please review, including all the comments and rationales and such. This is not a trivial change, so I expect some extended testing will be necessary.

### DIFF
--- a/GameData/Kerbalism/Profiles/Default.cfg
+++ b/GameData/Kerbalism/Profiles/Default.cfg
@@ -890,6 +890,14 @@ Profile
     maxAmount = 10000
   }
 
+  // To support the pressure control
+  RESOURCE
+  {
+    name = Nitrogen
+    amount = 10000
+    maxAmount = 10000
+  }
+
   RESOURCE
   {
     name = Water
--- a/GameData/Kerbalism/Profiles/Default.cfg
+++ b/GameData/Kerbalism/Profiles/Default.cfg
@@ -875,6 +875,13 @@ Profile
     amount = 10000
     maxAmount = 10000
   }
+
+  RESOURCE
+  {
+    name = Water
+    amount = 1500
+    maxAmount = 1500
+  }
 }
 
 
--- a/GameData/Kerbalism/Profiles/Default.cfg
+++ b/GameData/Kerbalism/Profiles/Default.cfg
@@ -789,7 +789,7 @@ Profile
     // Harvest time is 200 days, but in order to avoid micromanagement this greenhouse supports 0.5 Kerbal for 210 days.
     // Kerbals need 1258 food per 200 days, converted to 210 days that is 1320.9 food.
     crop_size = 660.45                  // amount of resource produced by harvests
-    crop_rate = 0.00000023148           // growth per 4 seconds when all conditions apply, a fully grown crop equals value of 1.0
+    crop_rate = 0.00000023148           // growth per second when all conditions apply, a fully grown crop equals value of 1.0
     ec_rate = 2.5                       // EC/s consumed by the lamp at max intensity
 
     light_tolerance = 400.0             // minimum lighting flux required for growth, in W/m^2
@@ -803,7 +803,7 @@ Profile
     INPUT_RESOURCE
     {
       name = Ammonia
-      rate = 0.00229                    // 15 units required per unit of crop, i.e. 200 days * 900 "4 seconds" per hour * 24 hours * rate == 15 * crop_size
+      rate = 0.00229                    // 15 units required per unit of crop, i.e. 200 days * 3600 seconds per hour * 6 hours * rate == 15 * crop_size
     }
 
     // Photosynthesis reaction: 2xH20 + CO2 -> O2 + CH20 + H20
@@ -811,7 +811,7 @@ Profile
     INPUT_RESOURCE
     {
       name = Water
-      rate = 0.0000765                  // 0.5 units required per unit of crop, i.e. 200 days * 900 "4 seconds" per hour * 24 hours * rate == 0.5 * crop_size
+      rate = 0.0000765                  // 0.5 units required per unit of crop, i.e. 200 days * 3600 seconds per hour * 6 hours * rate == 0.5 * crop_size
     }
 
     INPUT_RESOURCE
--- a/GameData/Kerbalism/Profiles/Default.cfg
+++ b/GameData/Kerbalism/Profiles/Default.cfg
@@ -770,8 +770,16 @@ Profile
     name = Greenhouse
 
     crop_resource = Food                // name of resource produced by harvests
-    crop_size = 2500.0                  // amount of resource produced by harvests
-    crop_rate = 0.00000023148           // growth per-second when all conditions apply
+    // Based on design targets from Prototype Lunar Greenhouse (see https://www.ag.arizona.edu/lunargreenhouse/MidReviews.htm):
+    // Design targets are on slide 18 of https://www.ag.arizona.edu/lunargreenhouse/Documents/2012-07-20_01_Giacomelli.pdf
+    // Prototype Lunar Greenhouse geometry can be found on slide 14 of https://www.ag.arizona.edu/lunargreenhouse/Documents/2012-07-20_01_Giacomelli.pdf
+    // Much larger design for producing food for 6 people can be found here (values not used): https://www.degruyter.com/downloadpdf/j/opag.2017.2.issue-1/opag-2017-0011/opag-2017-0011.pdf
+    // This Greenhouse is assumed to have 20 m^3 volume dedicated to food production (all greenhouses in mod support files are calculated relative to this one).
+    // This Greenhouse is intended to support 0.5 Kerbal just like the Prototype Lunar Greenhouse.
+    // Harvest time is 200 days, but in order to avoid micromanagement this greenhouse supports 0.5 Kerbal for 210 days.
+    // Kerbals need 1258 food per 200 days, converted to 210 days that is 1320.9 food.
+    crop_size = 660.45                  // amount of resource produced by harvests
+    crop_rate = 0.00000023148           // growth per 4 seconds when all conditions apply, a fully grown crop equals value of 1.0
     ec_rate = 2.5                       // EC/s consumed by the lamp at max intensity
 
     light_tolerance = 400.0             // minimum lighting flux required for growth, in W/m^2
@@ -785,19 +793,19 @@ Profile
     INPUT_RESOURCE
     {
       name = Ammonia
-      rate = 0.00695                    // 37530 units required for crop
+      rate = 0.00229                    // 15 units required per unit of crop
     }
 
     INPUT_RESOURCE
     {
       name = Water
-      rate = 0.00023148                 // 1250 units required for crop
+      rate = 0.0000765                  // 0.5 units required per unit of crop
     }
 
     OUTPUT_RESOURCE
     {
       name = Oxygen
-      rate = 0.00463                    // 25% of oxygen required by 1 crew member
+      rate = 0.02758                    // 100% of oxygen required by 1 crew member, based on Prototype Lunar Greenhouse design targets (percentage is factor 2 larger than food)
     }
 
     OUTPUT_RESOURCE
--- a/GameData/Kerbalism/Profiles/Default.cfg
+++ b/GameData/Kerbalism/Profiles/Default.cfg
@@ -203,8 +203,8 @@ Profile
     modifier = _WaterRecycler
     input = ElectricCharge@0.0446
     input = WasteWater@0.0001115
-    output = Water@0.0000892
-    output = Waste@0.0000223
+    output = Water@0.0001115
+    output = Waste@0.0000007433 // Waste is 150 times more dense than the density difference between Water and WasteWater
     dump = Waste
   }
 
@@ -216,8 +216,10 @@ Profile
     modifier = _WasteProcessor
     input = ElectricCharge@0.065556
     input = Waste@0.00010926
-    output = Ammonia@0.05463
-    dump = Waste
+    // According to http://ceadserv1.nku.edu/longa/haiti/kids/feces_value.html human waste contains approximately 13% nitrogen based on dry weight
+    // Human waste is considered to the dominant source of Waste
+    // Waste is 975.3 times more dense than Ammonia
+    output = Ammonia@.01385296614
   }
 
   // convention: 1 capacity = enough to compress output of 1 crew member
--- a/GameData/Kerbalism/Profiles/Default.cfg
+++ b/GameData/Kerbalism/Profiles/Default.cfg
@@ -776,7 +776,7 @@ Profile
     // Design targets are on slide 18 of https://www.ag.arizona.edu/lunargreenhouse/Documents/2012-07-20_01_Giacomelli.pdf
     // Prototype Lunar Greenhouse geometry can be found on slide 14 of https://www.ag.arizona.edu/lunargreenhouse/Documents/2012-07-20_01_Giacomelli.pdf
     // Much larger design for producing food for 6 people can be found here (values not used): https://www.degruyter.com/downloadpdf/j/opag.2017.2.issue-1/opag-2017-0011/opag-2017-0011.pdf
-    // This Greenhouse is assumed to have 20 m^3 volume dedicated to food production (all greenhouses in mod support files are calculated relative to this one).
+    // This Greenhouse is assumed to have 24 m^3 volume dedicated to food production (all greenhouses in mod support files are calculated relative to this one).
     // This Greenhouse is intended to support 0.5 Kerbal just like the Prototype Lunar Greenhouse.
     // Harvest time is 200 days, but in order to avoid micromanagement this greenhouse supports 0.5 Kerbal for 210 days.
     // Kerbals need 1258 food per 200 days, converted to 210 days that is 1320.9 food.
@@ -795,7 +795,7 @@ Profile
     INPUT_RESOURCE
     {
       name = Ammonia
-      rate = 0.00229                    // 15 units required per unit of crop
+      rate = 0.00229                    // 15 units required per unit of crop, i.e. 200 days * 900 "4 seconds" per hour * 24 hours * rate == 15 * crop_size
     }
 
     // Photosynthesis reaction: 2xH20 + CO2 -> O2 + CH20 + H20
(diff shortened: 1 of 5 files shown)


madman2003:
@MoreRobustThanYou I will for sure, already noticing some quirks that need attention before this can be merged

PiezPiedPy:
@MoreRobustThanYou  I think Waste for the compressor and Incinerator should be changed to Rubbish, as with these changes there is ambiguity between the different types of waste for the processes.
I will change modify the processes and add the resources if you have not got a problem with the changes.

madman2003:
@PiezPiedPy The ambiguity was always there, using Waste both for nitrogen recovery, as well as for other purposes (shielding and burning). I do agree that a multi stage waste setup is beneficial. I personally won't get around to a deeper round of playtesting until a day from now.

madman2003:
I will try to think if it is possible to incorporate direct usage of WasteAtmosphere instead of CO2, because using scrubbers is kind of silly when you have plants. The issue is that excess CO2 for food production has to be supplied without poisoning people on-board.

PiezPiedPy:
> The ambiguity was always there

@madman2003 Your right there, I'll make the changes, makes more sense for the processes.

GameData/Kerbalism/Profiles/Default.cfg
```diff
@@ -785,25 +803,33 @@ Profile
     INPUT_RESOURCE
     {
       name = Ammonia
-      rate = 0.00695                    // 37530 units required for crop
+      rate = 0.00229                    // 15 units required per unit of crop, i.e. 200 days * 900 "4 seconds" per hour * 24 hours * rate == 15 * crop_size
     }
 
+    // Photosynthesis reaction: 2xH20 + CO2 -> O2 + CH20 + H20
+    // Note that CH20 is an intermediate carbohydrate, energy that goes into the food, in the actual food it can be any CxHyOz compound
     INPUT_RESOURCE
```
PiezPiedPy: Is H20 a Hydrogen molecule consisting of 20 bonded Hydrogen atoms or did you mean water H2O?

GameData/Kerbalism/System/Resources.cfg
```diff
@@ -5,7 +5,7 @@
 RESOURCE_DEFINITION
 {
   name = Atmosphere
-  density = 0.001251          // 1 m^3 of nitrogen at STP
+  density = 0.000001251          // nitrogen at STP in units metric tonne/L or Mg/dm^3 (factor 1 million smaller than typical units kg/m^3)
```
PiezPiedPy: I don't think changing the density is a good idea for the resources because all densities used in Kerbalism are kg/m^3

GameData/Kerbalism/Profiles/Default.cfg
```diff
@@ -170,7 +178,7 @@ Profile
   {
     name = non-regenerative scrubber
     modifier = _NonRegenScrubber
-    input = WasteAtmosphere@0.00004       // capacity start at 2, reach 1 in 6h, then drop below
+    input = WasteAtmosphere@0.04       // capacity start at 2, reach 1 in 6h, then drop below
```
PiezPiedPy: Best to leave these types of values based on their previous densities

GameData/Kerbalism/Profiles/Default.cfg
```diff
@@ -203,8 +211,8 @@ Profile
     modifier = _WaterRecycler
     input = ElectricCharge@0.0446
     input = WasteWater@0.0001115
-    output = Water@0.0000892
-    output = Waste@0.0000223
+    output = Water@0.0001115
```
PiezPiedPy: This value is incorrect, you have not removed the waste from it

PiezPiedPy:
I've got a question, why is there a WasteWater Output and what does it consist of, the only waste water I can think of a greenhouse producing would be via evaporation which is only Water, would it not be better for the greenhouse to output Waste i.e. a HydroCarbon+Nitrogen compost type of waste?

PiezPiedPy:
@madman2003 Are you aware that the volumes used in KSP are 5 Litres/Unit 😉

GameData/Kerbalism/Profiles/Default.cfg
madman2003: That's a typo, it should be H2O

GameData/Kerbalism/Profiles/Default.cfg
madman2003: WasteWater has higher density than Waste, you cannot just split the volume assuming they are equal density (check community resource pack for densities), the assumption here is that waste dissolved into the Water to form WasteWater and the volume of Water didn't decrease. What would be your proposal?

madman2003:
@PiezPiedPy 

I didn't design the WasteWater output for the greenhouses on purpose, i just left it alone mostly. I know it isn't clean water, considering space systems are hydroculture (read: plants sitting in water instead of soil) based it should be mostly Water with residual nutrients that weren't absorbed. But we can consider that water flow to be be closed loop as well, so that we only have to supply water lost and absorbed nutrients. What are your thoughts?

I noticed something funky with Volumes indeed, but I noticed  a factor 4, not 5. But that could be the error you mention. Do you have a reference for this 5 Liters/Unit definition (because it sounds like a very unusual unit)?

If you make improvements, I'll obviously resolve the merge conflicts. Assuming they are smaller changes they are likely to go in earlier.

madman2003:
I did a test yesterday with a Water tank, and noted that every unit of water added 1 kg to the craft, which suggest that it is 1 liter. https://github.com/BobPalmer/CommunityResourcePack/blob/master/FOR_RELEASE/GameData/CommunityResourcePack/CommonResources.cfg#L525 shows density of 0.001 metric tonne/liter.

So I'm curious where this per 5 liter is coming from.
But an answer for this factor 4 i noticed in rates (everything became "per 4 seconds") would be really nice.

PiezPiedPy:
@madman2003 Densities are correct but volumes of the KSP stock tanks show as 5L/Unit see here: https://forum.kerbalspaceprogram.com/index.php?/topic/45231-what-is-the-unit-of-measure-for-rocket-fuel-it-does-not-appear-to-be-liters/
Kerbalism appears to have incorrect values in a lot of places, I plan to fix them all in the very near future.
ShotGunNinja seems to have assumed 1L/Unit that is probably why the weights of the water tanks show as 1L/Unit

PiezPiedPy:
@madman2003 The factor of 4 might be due to a Kerbal day is 6hrs and not 24hrs

PiezPiedPy:
@madman2003 Re: WasteWater, I think it would be better to keep the WasteWater closed loop for simplicity with the supplied water transferred into the Food content and the same with the Ammonia.

A friend of mine runs a small Aquaponics farm, his waste water consists of unused nutrients that are removed before the water goes back to the fish and the waste nutrients are used for growing algae that the fish eat, he has bacterial grow tanks between stages for doing things with the nutrients before they get to the plants or algae, blows my mind lol, I can allways ask him questions if needed.

GameData/Kerbalism/Profiles/Default.cfg
PiezPiedPy: Thought so 😉

madman2003:
@PiezPiedPy 
* I'll simplify the water management for greenhouses then
* Indeed stock units are per volume of 5, but all of the community units (which are not chosen by ShotGunNinja alone i suspect) seem to use volume of 1, so beware
* Good to know about the day length, that makes the comments a bit more straight forward, lol

PiezPiedPy:
@madman2003 Thanks for the tip regarding the Community units being 1L/Unit. I think it would be better to use 1L/Unit as it makes more sense but for the processes that output stock fuels I will use 5L/Unit for those specific outputs.

GameData/Kerbalism/Profiles/Default.cfg
PiezPiedPy: A change in density will alter the mass not volume, if the water content of the wastewater is removed you are left with the waste content.
Imagine you have 80 Litres of water and you dissolve 20 Litres of waste into it, you are going to have 100 Litres of waste water regardless of the materials having differing densities.

madman2003: @PiezPiedPy You are talking about mixing liquids, but suspending minor amounts of solids inside a liquid doesn't necessarily behave that way (just imagine your urine). The density increase of WasteWater compared to Water means at least the mass per volume unit is increasing. The approximation that I took was that the additional mass per unit of WasteWater compared to Water i converted to volume units of Waste based on the density of Waste. An approximation I admit, but treating a solid inside a liquid as two liquids of same density, where they add up based on volume also seems wrong. Preservation of mass is a much stronger criteria than preservation of volume.

madman2003:
Or make a single process that converts Waste in one go to all its components.

PiezPiedPy:
> Part of me wonders if we shouldn't be releasing water when we burn food, but I would need to look deeper for that.

@madman2003 I've been looking into that problem. Kerbal's outputs of Waste, Wastewater and CO2 compared to their consumption of food, water and O2 for breathing need work. So far we will maybe need to do something about how the rules work, as the rues only allow one input and output which causes a problem with modelling the Kerbal's Bio systems.
As one example, currently Kerbal's output of WasteWater is directly linked to the amount of Water they drink but in reality an additional 6.5% on average of the WasteWater they output should come from the burning of Food with the O2 they breath, which to simulate correctly needs rules to allow multiple inputs.

My calculations so far are (Kerbal's have basic Bio-systems so no complex amino acids etc. I think that's why they are green 😆)
```
Kerbal Pee Calculations
Input - Water H2O     87%  (so far this is 100% of the water that they drink)
Input - Food 2xNH3+C   6.5%
Input - Oxygen O2   6.5%  (extra water in the output comes from the Foods ammonia and O2)

Output - WasteWater -> (consisting of  Water H2O 93.5% + Urea CH4N2O 6.5%)
``` 
I'm calculating Kerbal's Waste output in a similar way.
I'll calculate actual rates using molar mass and densities but we still need to either modify the code that handles rules to accept multiple inputs or create an intermediate hidden process. We may also have to adjust the densities of Food, Waste and WasteWater in CRP to match final results.

Also as a by product of these calculations we will have a chemical composition for Food, Waste and WasteWater which will aid in fine tuning the Greenhouses.

PiezPiedPy:
> Regarding specializing Waste further, in order to do that we need a good proposal and get it into community resource pack.

When I created the Waste Incinerator process I used a chemical composition for the input Waste of H18C82 and used CRP's density for Waste to simulate all the snacks wrappers, also for the EC output I used BOE https://en.wikipedia.org/wiki/Barrel_of_oil_equivalent.
I think this would be a good starting point and as I suggested maybe call the resource Rubbish. I propose we use this in Kerbalism.
Whether @BobPalmer incorporates it or not into CRP is another matter.

madman2003:
@PiezPiedPy If at all possible, we should try to keep the carbon loops, the nitrogen loops and the water loops somewhat seperated. Emperical data is usually not presented like an all in one chemical formula. For example https://www.quora.com/How-much-water-does-a-person-lose-in-a-day-through-breathing gives some idea of water flows (we are not recovering water from the atmosphere), but it doesn't give the whole chemistry story. It's a different abstraction level, for good reason :-P

Part of me is tempted just to go with the partially unspec'ed nature of Waste and just recover both CO2 (see https://www.esa.int/Our_Activities/Space_Engineering_Technology/Melissa/Closed_Loop_Compartments for suggestion of recovering CO2 from solid waste) and nitrogen compounds from it. Because I don't believe we can tune everything in one go and community resource pack is a very slow moving target, so first time right is the way to go there

MoreRobustThanYou:
If someone feels like taking up the project, i'd love it if you could choose different crops to grow in the green house, each with different harvest size / grow time / light requirement / etc

PiezPiedPy:
@madman2003 Been thinking about what you posted, it will be better to adjust the output ratios of the Eating, Drinking and Breathing rules and reclaim the resources through the Water Recycler and Waste Processor rather than my way of using multiple inputs on the rules and then the processes.

madman2003:
@MoreRobustThanYou In reality you need a mixture of crops to keep people alive, and space people would in fact harvesting more often than once or twice per year. Taking away all your food in one go would totally wipe out the O2 producing capability and you'd die. And I'm not a big fan of making things more complex than needed. I just left the harvest mechanism until we go full blown bio regenerative system.

Right now the biggest blocker on this PR is that we're loosing too much water, when trying to produce food. Just need to figure out where water loss is going (i.e. check the website i used to model the greenhouse, and hope to discover more insights).

@PiezPiedPy From a simplicity point yes, that only leaves the moisture in air problem. Not sure if and how to model that yet.

PiezPiedPy:
> Just need to figure out where water loss is going

@madman2003 I would assume the water is contained in the food, the waste processor does not output that water content yet it only recovers the ammonia.
Also if we are not going to go the route of multiple input rules as used in my previous post, then the water recycler could output approx. 6.5% more water too since a part of the waste water contains water from food.

PiezPiedPy:
@madman2003 I think the Water Recycler should output Water, CO2 and Ammonia rather than waste which also should have a water content that is extracted by the waste processor.
At the moment the waste the Kerbal's poop out is bone dry which must give them a hard time on the toilet 😬 I think they would appreciate us if we add some water content to it.

madman2003:
@PiezPiedPy Makes sense for the water recycler, but how do you rpropose to recovering things from solid waste? Which is not easily done with a chemical reactor.

Why not introduce a MoistAtmosphere (as counterpart to WasteAtmosphere) and recover from there? In real life you need to recover water lost to the atmosphere as well.

On top of that, you cannot recover 100% of urine in real life (https://www.space.com/34688-recycled-astronaut-pee-boosts-deep-space-travel.html suggests 75%) so do we want to tune that down as well, and how far?

PiezPiedPy:
@madman2003 The waste currently is solid, I'm proposing we add water to it by having a water output on the Waste Processor.

The processes in Kerbalism that mimic real life are 100% efficient, might as well do the same with the water and waste processors too.

PiezPiedPy:
We could simulate moisture reclamation in the scrubber since that is the process that handles the WasteAtmosphere by adding a Water output to it.
```
  // convention: 1 capacity = enough to scrub output of 1 crew member
  Process
  {
    name = scrubber
    modifier = _Scrubber
    input = ElectricCharge@0.025
    input = WasteAtmosphere@0.00002
    output = Water@ // Need to calculate a rate to simulate moisture reclamation.
    output = CarbonDioxide@0.02
    dump = true
  }

```

madman2003:
@steamp0rt Can you fix CI, it doesn't trigger anymore after your name change?

madman2003:
@PiezPiedPy: The scrubber is a separate piece of hardware. plus WasteAtmosphere is modeled after CO2, water recovery does not come for free on a small space ship, you need a station sized thing for it, putting in the scrubber makes it too easy (first piece of hardware that goes on any lander)

madman2003:
P.S. I do appreciate the feedback :wink:

PiezPiedPy:
@madman2003 No probs 😉

A rule for MoistAtmosphere it is then, plus a process to reclaim the water. I think we should add the process to command pods and also the chemical processors, what do you think ?
I will add them in and submit in a separate PR after I calculate a moisture content and density, any ideas?

madman2003:
Look at this: https://www.quora.com/How-much-water-does-a-person-lose-in-a-day-through-breathing
It seems like 900 ml goes into the atmosphere, just use the density of water, just like WasteAtmosphere is just CO2. All you want to track is how much there is in the atmosphere, no actual atmosphere density or composition.

Moisture recovery usually isn't perfect in a single pass, but over the whole you keep filtering air and the water didn't go anywhere, so for simplicity sake model it like a perfect process. ISS has two systems (one US, other Russian) that process moisture, and a crew of 6. So don't go overkill on processing for a dozen people :-P

madman2003:
Do decide if you want to recover WasteWater or directly Water.
In the meantime I'm going to check I'm overproducing ammonia

PiezPiedPy:
Will be simpler to reclaim the water, just been reading about the ISS Atmospheric Condensate Water Regeneration System it converts the condensate into potable water.

madman2003:
@PiezPiedPy When astronouts wash themselves water also floats around the cabin (see for example https://www.youtube.com/watch?v=tDbbJWKKQu0), we haven't even considered those factors, so I'd avoid the 1.2L value ;-)

PiezPiedPy:
@madman2003 Well spotted, I'll use a lower value ;)

madman2003:
I did some checking and with the current PR i can sort of configure a working station again, nitrogen must be supplied, but can last the lifetime of the station basically. Water and CO2 must be supplied also far more frequently (once a year given a big enough supply, or once every few years), CO2 for food production, Water for for oxygen production, food production, and urine water loss. Food production for large group of people is non-trivial, basically two kerbalism greenhouses are needed per person. So 2.5m diameter by 20 meters.

Only thing missing is electrolysis with hydrogen priority, which I will still add.

@steamp0rt and @PiezPiedPy anything you want to have checked before accepting this PR?
I'm assuming it will be tuned again later, but this is a pretty big chunk that I don't want to have hanging in a PR forever.

PiezPiedPy:
Looks fine to me, a much better system than it previously was, thanks for the work you've put into this.
I think it deserves to be merged.

PiezPiedPy:
@madman2003 I used ModuleManager to change the parts when I did the USI support config but I have no idea where the Configure node is spawned from as it is not in the Part cfg files. I think it may be set in code somehow in the `Configure.cs` file.

This is how I did it
```
@PART[kerbalism-container-radial-*]:NEEDS[UmbraSpaceIndustries,Kerbalism,ProfileDefault]:AFTER[UmbraSpaceIndustries]
{
	@MODULE[Configure]
	{
		SETUP
		{
			name = Liquid Hydrogen
			desc = Store <b>Liquid Hydrogen</b>.
			tech = advFuelSystems

			RESOURCE
			{
				name = LqdHydrogen
				amount = 36.26
				maxAmount = 36.26
				@amount *= #$../../../ContainerVolume$
				@maxAmount *= #$../../../ContainerVolume$
			}
		}
	}
}
```

PiezPiedPy:
@madman2003 see line 761 in the `default.cfg` file, you can change the amount there 😉

PiezPiedPy:
@madman2003 As a suggestion update the `CHANGELOG.md` file with your changes and credit yourself.

GameData/Kerbalism/Profiles/Default.cfg
```diff
@@ -105,6 +113,7 @@ Profile
     input = Water
     output = WasteWater
     rate = 0.605 // 2.42 Kg per-day
+    ratio = 1.000 // Every unit of Water becomes a unit of WasteWater, extra mass comes from inside body
```
steamp0rt: rate AND ratio? If you test it and this works, then this is ok.

GameData/Kerbalism/Profiles/Default.cfg
```diff
@@ -48,6 +48,14 @@ Profile
     refill_message = $VESSEL oxygen reserves restored@<i>The crew is taking a breather</i>
   }
 
+  Supply
+  {
+    resource = CarbonDioxide
```
steamp0rt: This may get spammy.

madman2003: It's not really, either your ship doesn't store CO2 and you don't get any messages about CO2 (even if you convert the CO2 on the spot into water again). Or you do store CO2 and you care. Or you are in the steady state of empty, which produces the message once, and then it stops. Theoretically you can have non-steady state consumption that frequently crosses the threshold, but that would affect every other resource as well, and is pretty rare.

madman2003:
@steamp0rt Satisfied with answers?

GameData/Kerbalism/Profiles/Default.cfg
madman2003: Not to mention that listing as Supply is neccesary for it to show up in the planner.

steamp0rt:
Yes

madman2003:
@steamp0rt I'm guessing you are waiting to test yourself?

steamp0rt:
I'm going to take your word on it.

PR #11, opened 2018-05-13
Humidity Control, Rules, Processes and a Life Support Unit

PiezPiedPy:
## Humidity Control with added moisture for those days when you want to write your name on the space station windows.

As talked about in PR #10 I have created a Humidity Control that enables Kerbal's to output Water via breathing.

Kerbal's breathe out 850mL of Water per day, this water is from the kerbal's bio system that internally processes food, the food is not taken as an input resource in the humidity rule as the food is accounted for via the eating rule as normal.

A new hidden resource `MoistAtmosphere` has been added to track the humidity.

To achieve the desired results I had to add two new parameters to the Rules system, these parameters can also be used with any rule that needs the extra functionality. The new parameters are:
* `input_threshold` When the input resource reaches this percentage of its max capacity, degeneration is triggered rather than the default of a zero resource amount triggering the degeneration [range 0 to 1].
* `output_only` If true and an input resource exists, only monitor the input resource and do not consume it.

Eventually I will move Humidity Control into the Habitat system but for now I feel this is adequate.

I have also added a Life Support Unit to the Parts folder, originally the parts cfg file was passed to me by a member on the KSP Forum a couple of years ago but since then I have added all the Life Support processes, it includes all the things needed for Life Support that are on the pods but in its own part that can be fitted when pods don't have enough slots for required needs, especially handy for space stations and bases.


PiezPiedPy:
@steamp0rt Yes I've tested it 😛. Now I'm of to 🛏

GameData/Kerbalism/Parts/LifeSupport/kerbalism-lifesupportmodule.cfg
```diff
@@ -0,0 +1,189 @@
...
+    type = ProcessController
+    title = ECLSS
+    redundancy = Life Support
+    repair = true // anyone
+    mtbf = 72576000 // 8y
+    extra_cost = 2.5
+    extra_mass = 0.1
+  }
+  
+  MODULE
```
madman2003: All this stuff, can this not be configured with the same config that applies it to every pod? This smell like duplication.

GameData/Kerbalism/Parts/LifeSupport/kerbalism-lifesupportmodule.cfg
```diff
@@ -0,0 +1,189 @@
...
+    capacity = 2
+    running = true
+  }
+
+  @MODULE[ProcessController],*
+  {
+    @capacity *= 1.67
+  }
+
+  MODULE
```
madman2003: Same here, this duplicated config is begging to be be forgotten in a future update.

GameData/Kerbalism/Profiles/Default.cfg
```diff
@@ -127,13 +127,30 @@ Profile
     relax_message = $ON_VESSEL$KERBAL is breathing again
   }
 
+  Rule
+  {
+    name = humidity
+    input = MoistAtmosphere
+	output = MoistAtmosphere
```
madman2003: check your indentation

GameData/Kerbalism/Profiles/Default.cfg
```diff
@@ -127,13 +127,30 @@ Profile
     relax_message = $ON_VESSEL$KERBAL is breathing again
   }
 
+  Rule
+  {
+    name = humidity
+    input = MoistAtmosphere
```
madman2003: What does it mean to have the same input and output? Does this actually produce MoistAtmosphere?

GameData/Kerbalism/Profiles/Default.cfg
```diff
@@ -196,6 +213,17 @@ Profile
...
   }
 
+  // convention: 1 capacity = enough to reclaim moisture output of 1 crew member
+  Process
+  {
+    name = humidity controller
+    modifier = _HumidityController
+    input = ElectricCharge@0.037
+    input = MoistAtmosphere@0.002879122904 // 50Ltr per-day @100% humidity, contains 850 mL of Water per-day
+    output = Water@0.0000393519
```
madman2003: I'd appreciate rationales for the ratios ;-)

GameData/Kerbalism/System/Resources.cfg
```diff
@@ -26,6 +26,18 @@ RESOURCE_DEFINITION
   volume = 1
 }
 
+RESOURCE_DEFINITION
+{
+  name = MoistAtmosphere 
+  density = 0.000804          // 1 m^3 of saturated water vapor at STP , 100% humidity
```
madman2003: A specific reason to model it like 100% humidity air? From what I understand this implies conversion from Atmosphere to MoistAtmosphere (including messing with the nitrogen cycle involved with atmosphere generation). WasteAtmosphere is just CO2, why not make this just Water (density)?

src/Profile/Rule.cs
```diff
@@ -16,20 +16,22 @@ public Rule(ConfigNode node)
 			interval = Lib.ConfigValue(node, "interval", 0.0);
 			rate = Lib.ConfigValue(node, "rate", 0.0);
 			ratio = Lib.ConfigValue(node, "ratio", 0.0);
-			degeneration = Lib.ConfigValue(node, "degeneration", 0.0);
-			variance = Lib.ConfigValue(node, "variance", 0.0);
+            input_threshold = Lib.ConfigValue(node, "input_threshold", 0.0);
```
madman2003: What happened with indenting here? Check rest of file too.

GameData/Kerbalism/Profiles/Default.cfg
```diff
@@ -127,13 +127,30 @@ Profile
     relax_message = $ON_VESSEL$KERBAL is breathing again
   }
 
+  Rule
+  {
+    name = humidity
+    input = MoistAtmosphere
+	output = MoistAtmosphere
+    rate = 0.002879122904 // 50Ltr per-day @100% humidity contains 850 mL of Water per-day per Kerbal, (humans breath is at 100% humidity)
+    output_only = true // monitor input do not consume it
```
madman2003: Make a note maybe the assumption is here that kerbals will burn fat reserves when not getting food, otherwise this thing makes no sense.

madman2003: Can output_only not be implicit by not configuring an input?

GameData/Kerbalism/Parts/LifeSupport/kerbalism-lifesupportmodule.cfg
PiezPiedPy: @madman2003 That's a point, never thought of that, I just pulled it out of my personal changes folder as a last thought and dropped it into the branch, gave it a quick test to make sure the Humidity bit I dropped into it worked and submitted the PR.
I'll change it later today as I'm about to go to bed 😉

madman2003:
Good work, I gave you comments because you deserve them, not out of dislike ;-)

src/Profile/Rule.cs
PiezPiedPy: looks fine in Visual studio and notepad++

madman2003: original code was tab indented, new code is space ind

GameData/Kerbalism/Profiles/Default.cfg
madman2003: At the very least this is worthy of a comment so the poor souls who read this can make sense of it ;-)

GameData/Kerbalism/System/Resources.cfg
PiezPiedPy: Kerbalisms Atmosphere, Nitrogen or CO2 has not been used in any of the calculations nor used in any outputs or inputs to the rules or processes used by Humidity, there is no conversion between them.

GameData/Kerbalism/Profiles/Default.cfg
PiezPiedPy: Basically there is 850mL of Water in 50Ltr of saturated (100% humidity) MoistAtmosphere. Which is the same amount a Kerbal outputs in a day.
Input rate of MoistAtmosphere is calculated using:
* density of saturated Air (100% Humidity) of (8.040000e-04) (this has been multiplied by 1e3 actual real world density is 8.040000e-07 only reason I multiplied is it looked better in game otherwise input is 2.8 Units rather than 0.0028 but anyway its not so important and can be changed).
* grams/KSPUnit using density gives us 804grams/Ltr (Heavy Metal I mean Air lol).
* amount of water in a Litre of air used in calculations is 0.017mL (in real world this is approx. 100% saturation).
* 850mL of Water divided by 0.017mL = 50 Ltrs  (850mL of Water is in 50Ltrs of saturated air).
* 50Ltr divided by 804grams/Ltr = 62.18905473 Ltr of MoistAtmosphere 
* 62.18905473 Ltr / 21600 seconds (a Kerbal day of 6hrs) = 0.002879122904

Water output was even easier:
* density of water (duhh) 😛 1.0e-3 = 1000 grams/Ltr or (grams/KSPUnit)
* 850mL divided by 1000mL = 0.85 Ltr of Water
* 0.85 Ltr / 21600 seconds (a Kerbal day of 6hrs) = 0.0000393519

GameData/Kerbalism/Parts/LifeSupport/kerbalism-lifesupportmodule.cfg
PiezPiedPy: nothing to see here, move along 🤣

src/Profile/Rule.cs
PiezPiedPy: ahh makes sense, @steamp0rt changed all the spacing to tabs not so long ago and my visual studio is set as spaces >.<

PiezPiedPy:
@steamp0rt Hold the merge of this till I change a few things, I'll let you know when its ready.

steamp0rt:
Ok

GameData/Kerbalism/Profiles/Default.cfg
PiezPiedPy: Comments added 😉

GameData/Kerbalism/Profiles/Default.cfg
PiezPiedPy: indents fixed

GameData/Kerbalism/Parts/LifeSupport/kerbalism-lifesupportmodule.cfg
PiezPiedPy: This file will have to stay as it is because the pod configuration in the default profile only works on parts with a crew capacity which this part does not have.
Only other option is to move the data out of this file and copy it into the default profile but that still leaves the problem of the data appearing to be duplicated.
Unless you know of another workaround ?

PiezPiedPy:
@steamp0rt This is ready to be merged unless you have any changes you would like to add.

steamp0rt:
Will do it in morning

GameData/Kerbalism/Parts/LifeSupport/kerbalism-lifesupportmodule.cfg
madman2003: This has the added benefit of allowing people to change it in their profile.

PiezPiedPy: @madman2003 I had the same thoughts but didn't want to bloat the default profile any more than needed, I'll have a think over it, I am more towards moving it to the default profile though, so that is probably what I will end up doing.

PR #12, opened 2018-05-13
Ui scaling

madman2003:
Just my playground for now. Using your jenkins job to as an easy way to get a build.
Once the other PR has gone through, and I'm sure this works I'll poke you ;-)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@
 
 ### Changes since the last release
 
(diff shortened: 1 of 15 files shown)


PiezPiedPy:
@madman2003 have you got a development environment ?

madman2003:
@PiezPiedPy My workflow has always been to change on my linux box, and then download on my windows box, but for the next thing I will invest in setting up a mono compiler setup and transfer over the network. I did notice this is causing a bit too much spam.

steamp0rt:
mono-complete

PR #15, opened 2018-05-15
UI Scale tweeks and fixes

PiezPiedPy:
Added `UIPanelWidthScale` user setting, allows the user to have a separate scale for the width of the UI windows.
* Reason I added this feature is that after I applied a UI scale, the InFlight Monitor panel was partially underneath my KER window. I had a feeling users would probably have the same problem, so I came up with this setting to change the width scale of the panels independently of the UI Scale.

Scaled the string lengths in the windows to stop overlapping.

Scaled the Alert Messages.

InFlight Monitor panels are now the same size to help stop overlapping.

Space Centre Monitor panels for science data , message log and unselected vessels are now wider than the other panels to show more information.

Science data and message log popout windows are now wider than the other popout windows again to show more information.

Popout windows no longer show a scroll bar due to their height not scaling enough.

All tooltips now have correct alignment of data.

Clicking the middle mouse button on the popout menu will now close the popout window if it is already open.

Message Log window still needs some work and certain scales still upset the tooltips but that can be addressed in a later PR.

I will post some pictures 😉

### DIFF
--- a/src/UI/FileManager.cs
+++ b/src/UI/FileManager.cs
@@ -43,7 +43,7 @@ namespace KERBALISM
 			{
 				string filename = pair.Key;
 				File file = pair.Value;
-				render_file(p, filename, file, drive, in_monitor && Lib.IsFlight());
+				render_file(p, filename, file, drive, short_strings && Lib.IsFlight());
 			}
 			if (drive.files.Count == 0) p.content("<i>no files</i>", string.Empty);
 
@@ -53,7 +53,7 @@ namespace KERBALISM
 			{
 				string filename = pair.Key;
 				Sample sample = pair.Value;
-				render_sample(p, filename, sample, drive, in_monitor && Lib.IsFlight());
+				render_sample(p, filename, sample, drive, short_strings && Lib.IsFlight());
 			}
 			if (drive.samples.Count == 0) p.content("<i>no samples</i>", string.Empty);
 		}
--- a/src/UI/FileManager.cs
+++ b/src/UI/FileManager.cs
@@ -67,9 +67,9 @@ namespace KERBALISM
 			string exp_label = Lib.BuildString
 			(
 			  "<b>",
-			  Lib.Ellipsis(exp.name, Styles.ScaleStringLength(short_strings ? 24 : 40)),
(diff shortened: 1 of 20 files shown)


PiezPiedPy:
ui scale

src/UI/FileManager.cs
```diff
@@ -88,7 +89,7 @@ static void render_file(Panel p, string filename, File file, Drive drive)
 		}
 
 
-		static void render_sample(Panel p, string filename, Sample sample, Drive drive)
+		static void render_sample(Panel p, string filename, Sample sample, Drive drive, bool in_monitor)
```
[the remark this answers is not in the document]
madman2003: also applies to other instances

src/UI/Styles.cs
```diff
@@ -144,12 +144,15 @@ public static float ScaleFloat(float val)
...
 		}
 
-		// Increasing font size does not affect width as much as height. To avoid excessively wide UIs we scale differently.
-		// Scaling by 80% is simply an emperical factor that is intended to leave the default UI width of 100% scaling intact,
-		// while having a sensible width at high scale value (e.g. 170%).
+		// Increasing font size does not affect width as much as height. To avoid excessively wide UIs we scale with UIPanelWidthSScale rather than	UIScale
 		public static float ScaleWidthFloat(float val)
 		{
-			return val * (1.0f + (Settings.UIScale * GameSettings.UI_SCALE * GameSettings.UI_SCALE_APPS - 1.0f) * 0.8f);
+			return val * (1.0f + (Settings.UIPanelWidthScale * GameSettings.UI_SCALE * GameSettings.UI_SCALE_APPS - 1.0f));
```
madman2003: you simplify this by removing the 1.0f values, they don't do anything anymore

madman2003:
Good work, left two minor comments.

src/UI/Styles.cs
PiezPiedPy: thanks ;)

PiezPiedPy:
@madman2003 Thanks for the heads up, I'll look into it

PiezPiedPy:
@madman2003 Just in case you are wondering, the string lengths for the titles etc are scaled with the method `ScaleStringLength` in the `Styles.cs` file

PR #19, opened 2018-05-21
Mostly SSPX fixes

madman2003:
See changelog ;-)


src/Modules/GravityRing.cs
```diff
@@ -36,30 +37,73 @@ public override void OnStart(StartState state)
...
 
 		public void Update()
 		{
 			// update RMB ui
 			Events["Toggle"].guiName = deployed ? Localizer.Format("#KERBALISM_Generic_RETRACT") : Localizer.Format("#KERBALISM_Generic_DEPLOY");
-			Events["Toggle"].active = deploy.Length > 0 && !deploy_anim.playing() && !waitRotation && ResourceCache.Info(vessel, "ElectricCharge").amount > ec_rate;
+			Events["Toggle"].active = (deploy.Length > 0) && (part.FindModuleImplementing<Habitat>() == null) && !deploy_anim.playing() && !waitRotation && ResourceCache.Info(vessel, "ElectricCharge").amount > ec_rate;
 		}
 
 		public void FixedUpdate()
```
PiezPiedPy: Might be better for performance issues to move code in the `FixedUpdate()` method into the `Update()` method.
`FixedUpdate()` should only be used for physics calculation math or time sensitive data that ***MUST*** be calculated every physics frame. Code in `FixedUpdate()` should be kept as short and optimized as possible. (see the `Profiler.cs` file and note the code that is called in `FixedUpdate()` is the code used for detecting the length of execution times only). 

A lot of people misunderstand the differences of the uses of these two methods and dump everything in `FixedUpdate()` because it is called every physics frame. 😉

PiezPiedPy: One day soon I will be going through all the files in Kerbalism and moving code that is not physics frame sensitive from `FixedUpdate()` to `Update()`

PiezPiedPy:
Nice work 😎

PR #21, opened 2018-05-26
Harvester balance

madman2003:
Bump up the water harvesting rates a bit.
Counteract the nerf that I implicitly did on harvesters in general.
Fix a bug (that I introduced) where harvesting only starts after saving game.
Don't waste energy on harvesting when full.


madman2003:
@robot-aquila I'll leave it up to you if you have water production proposals beyond this, keep in mind that Kerbalism is not sci-fi, so you probably need bigger hardware for very large scale water harvesting.

madman2003:
FYI In your 3% abundance scenario on minnus performance would be factor 1.5-2.0 compared to release 1.5.

robot-aquila:
Thank you. I will back with proposal document soon. I'm sure there is possible to provide a good balance between realistic numbers and use cases.

steamp0rt:
@robot-aquila I'm gonna go ahead and merge this, but your proposal is still welcome

Issue #9, opened 2018-05-06
ContractConfigurator issue still persists


Gordon-Dry:
I just copy this issue
https://github.com/ScienceFun/Kerbalism-Continued/issues/9
to not be forgotten, because it still exists.

The ModuleManager.ConfigCache will never be created because:
```
[ERR 17:31:32.958] [ModuleManager] Exception while processing node : Kerbalism/Support/ContractConfigurator/@CONTRACT_TYPE[*]

[EXC 17:31:32.961] NullReferenceException: Object reference not set to an instance of an object
	ConfigNode.CopyToRecursive (.ConfigNode node, Boolean overwrite)
	ConfigNode.CopyToRecursive (.ConfigNode node, Boolean overwrite)
	ConfigNode.CopyTo (.ConfigNode node, System.String newName)
	ModuleManager.MMPatchLoader.ModifyNode (ModuleManager.Collections.ImmutableStack`1 original, .ConfigNode mod, PatchContext context)
	ModuleManager.PatchApplier.ApplyPatches (System.String stage, IEnumerable`1 patches)
	UnityEngine.Logger:LogException(Exception)
	ModuleManager.Logging.UnityLogger:Exception(String, Exception)
	ModuleManager.Logging.ModLogger:Exception(String, Exception)
	ModuleManager.Logging.ExceptionMessage:LogTo(IBasicLogger)
	ModuleManager.<ProcessPatch>d__33:MoveNext()
	UnityEngine.SetupCoroutine:InvokeMoveNext(IEnumerator, IntPtr)
[ERR 17:31:32.961] [ModuleManager] Processed node was
Kerbalism/Support/ContractConfigurator/@CONTRACT_TYPE[*]
  @CONTRACT_TYPE[*]
  {
    @PARAMETER[*]:HAS[#partModule[ModuleScienceLab]]
    {
      @partModule = Laboratory
    }
    @PARAMETER[*]
    {
      @PARAMETER[*]:HAS[#partModule[ModuleScienceLab]]
      {
        @partModule = Laboratory
      }
      @PARAMETER[*]
      {
        @PARAMETER[*]:HAS[#partModule[ModuleScienceLab]]
        {
          @partModule = Laboratory
        }
        @PARAMETER[*]
        {
          @PARAMETER[*]:HAS[#partModule[ModuleScienceLab]]
          {
            @partModule = Laboratory
          }
        }
      }
    }
[...]

PR #25, opened 2018-05-27
ContractConfigurator support changes

PiezPiedPy:
Removed all now unused Antenna part references since the Signals feature was removed. Also moved the Needs sections to root.

Fixes the bug raised in Issue #9

Also bumped version to 1.6.1


Gordon-Dry:
Well, I should give it a try now.

PR #35, opened 2018-05-30
Resolve IDE messages

PiezPiedPy:
* Use latest available C# version and pattern matching
 * IDE0018 Message "Variable declaration can be inlined" suppressed because variable declaration inlining does not work on versions of Visual Studio older than VS2017
 * Analyzer references added to solve the IDE1003 errors

 * Add access modifiers to solve IDE0044 messages
 * Simplified object initialization to solve IDE0017 messages
 * Optimizations to remove the last of the IDE messages.


PiezPiedPy:
@steamp0rt  Wow - that was quick 😃

Issue #40, opened 2018-06-02
CLS support config has old part names in it


Gordon-Dry:
GameData\Kerbalism\Support\CLS.cfg:
```
@PART[Greenhouse]:HAS[!MODULE[ModuleConnectedLivingSpace]]:NEEDS[ConnectedLivingSpace]
{
  MODULE
  {
    name = ModuleConnectedLivingSpace
    passable = true
  }
}


@PART[GravityRing]:HAS[!MODULE[ModuleConnectedLivingSpace]]:NEEDS[ConnectedLivingSpace]
{
  MODULE
  {
    name = ModuleConnectedLivingSpace
    passable = true
  }
}
```

I guess it should be _kerbalism-greenhouse_ instead of _Greenhouse_
and _kerbalism-gravityring_ instread of _GravityRing_, or am I wrong?

Gordon-Dry:
see
https://github.com/steamp0rt/Kerbalism/issues/39

PiezPiedPy:
I'll check ;)

PiezPiedPy:
this need closing, fixed with PR #42

Gordon-Dry:
But I admit that on decoupling this still happens.

Issue #39, opened 2018-06-02
Supply containers got a "hole" in the center? Add CLS support.


Gordon-Dry:
I just see that I could stack them nicely, they have kinda hole in the center but no CLS support.

It would be fairly easy to add just the CLS support by patching, but also changing the texture of the top and bottom to show that they really are passable would be great.

btw I'm talking about the InlineContainers, at least the 125, 250 and 375 should get CLS support - the 0625 is too small...

Actually the 250 and 375 already got a texture on top/bottom that looks like they're passable.

PiezPiedPy:
Personally I have no idea how to modify the parts or textures, could you give details how to add the CLS support and I will include it.

Gordon-Dry:
New config, when I'm right just take it.
GameData\Kerbalism\Support\CLS.cfg
```
@PART[kerbalism-greenhouse]:HAS[!MODULE[ModuleConnectedLivingSpace]]:NEEDS[ConnectedLivingSpace]
{
  MODULE
  {
    name = ModuleConnectedLivingSpace
    passable = true
  }
}


@PART[kerbalism-gravityring]:HAS[!MODULE[ModuleConnectedLivingSpace]]:NEEDS[ConnectedLivingSpace]
{
  MODULE
  {
    name = ModuleConnectedLivingSpace
    passable = true
  }
}

@PART[kerbalism-container-inline-125]:HAS[!MODULE[ModuleConnectedLivingSpace]]:NEEDS[ConnectedLivingSpace]
{
  MODULE
  {
    name = ModuleConnectedLivingSpace
    passable = true
  }
}

@PART[kerbalism-container-inline-250]:HAS[!MODULE[ModuleConnectedLivingSpace]]:NEEDS[ConnectedLivingSpace]
{
  MODULE
  {
    name = ModuleConnectedLivingSpace
    passable = true
  }
}

@PART[kerbalism-container-inline-375]:HAS[!MODULE[ModuleConnectedLivingSpace]]:NEEDS[ConnectedLivingSpace]
{
  MODULE
  {
    name = ModuleConnectedLivingSpace
    passable = true
  }
}
```

PiezPiedPy:
Thanks

Gordon-Dry:
You could let the 125 out because the texture is not intended to be a passable part - but IMHO 1.25m is enough so the hole should be barely wide enough for a Kerbal head ... without helmet haha.

PiezPiedPy:
[...]

Issue #29, opened 2018-05-27
Signal loss during GameScene changes and SaveGame load


PiezPiedPy:
When changing GameScene's or loading a Save, Signals show as unconnected and then re-connect.

PiezPiedPy:
Maybe found the problem, seems to be due to Kerbalism.cs FixedUpdate method being executed before KSP has finished loading a scene.

I'll start to sort this out once PR's #25 and #30 have been merged.

Gordon-Dry:
Using the private build from PiezPiedBy this issue still occurs.

PiezPiedPy:
probably because I've not added any changes to fix it yet 😛

PiezPiedPy:
Fixed

PR #42, opened 2018-06-03
Fix signal lost/found on GameScene change

PiezPiedPy:
Signal lost/found during GameScene changes and SaveGame load annoying bug issue #29 fixed.

Also support for Connected Living Space bug issue #39 and #40 fixed. Textures for the models have not been altered though as I don't know how to do such things.

Part tags have been updated too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 
 ### Changes since the last release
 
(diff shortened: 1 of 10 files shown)


Gordon-Dry:
Ah!
> 	// hard-coded transmission rate and cost
> 	private const double ext_rate = 0.064;
> 	private const double ext_cost = 0.1;

PiezPiedPy:
@Gordon-Dry They are used as default rates if there are none defined for a part.

PR #48, opened 2018-06-09
Signal Overhaul

PiezPiedPy:
* Overhaul to transmitter use, planning and monitoring, data rates, signal strength, ec cost and targets now work.
 * Internal transmitters are separate from external transmitters and will only transmit telemetry and command control, they are also shown separately in the planner and are constantly powered unless you run out of ec, with the added benefit of loosing contact with DSN and subsequent control.
 * External transmitters will lose contact when retracted, break, or if you run out of ec rendering long distance comms, call home and your ability to transmit science to zero.
 * External transmitters will also stop using ec when retracted.
 * EVA suits now contain a small internal transmitter for transmitting telemetry and controlling remote probes and rovers etc.
 * All transmitters have had their ec usage changed to more realistic values and are also combinable. 

There is a minor drawback though, when changing scenes with the [ESC]Pause menu you may notice the target readout pointing to the wrong vessel and some signals that where previously dis-connected wrongly connecting back online, simply changing scene from for example the Space Centre to Tracking Station will solve all errors in the network.


Gordon-Dry:
I that like do.

Gordon-Dry:
This is not an issue!
Just a chitchat.

Checking this out and having to edit a private patch because of the new feature "Signal: Add transmitter to EVA suits and fixed typo's in profile" I just found out that Kerbals have no mass ...
I just had to add :HAS[!MODULE[KerbalEVA]] in my private patch to avoid 2 MM errors (for male and female Kerbals).

Gordon-Dry:
See
https://forum.kerbalspaceprogram.com/index.php?/topic/172400-131143-kerbalism-v160/&do=findComment&comment=3396878

PiezPiedPy:
@Gordon-Dry Thanks for testing 😃

Gordon-Dry:
Just referring to https://github.com/steamp0rt/Kerbalism/pull/54

Because I don't know how this commit and merge stuff works at the end (nothing Githubbish installed here) and just want to avoid some overriding ...

All this works now as it seems™.

Gordon-Dry:
So this should be merged now so you can proceed, @PiezPiedPy ?

PiezPiedPy:
@Gordon-Dry Yes

PR #79, opened 2018-06-27
Mtbf depends mass to have some variance

Gordon-Dry:
I feel good with my numbers now - if anybody has to yell about them this can be changed later, so I do my PR now.


GameData/Kerbalism/System/Reliability.cfg
```diff
@@ -4,688 +4,944 @@
...
+
+@PART[*]:HAS[@MODULE[Reliability],!MODULE[KerbalEVA]]:AFTER[Kerbalism]
+{
+	@MODULE[Reliability]:HAS[#mtbf[>145152000]],*
+	{
+		@mtbf = 145152000 // 16y highpass filter
+	}
+}
+
+@PART[*]:HAS[@MODULE[Reliability],!MODULE[KerbalEVA]]:FINAL
```
PiezPiedPy: Why `FINAL` ?  If it has to go very last then `AFTER[zzzKerbalism]` may be better

Gordon-Dry: Ok. I was not sure if a zzz thing even exists, I have not seen any in a quick search...

Gordon-Dry:
I'm fine with this.

PR #98, opened 2018-07-03
PlannerController title

PiezPiedPy:
Added a title parameter to the PlannerController that shows in the buttons text.


PiezPiedPy:
screenshot1

Makes more sense when a part has more than one module that has a planner button.

PR #101, opened 2018-07-03
radiation belt support patch for ExtraSolar 

Gordon-Dry:
The EXPERIMENT_DEFINITION is set, but
1. empty
2. commented out

If anybody dares to A) check if these presets are correct and B) add some neat texts ...


Gordon-Dry:
This post shows how they look:
https://forum.kerbalspaceprogram.com/index.php?/topic/130656-143-extrasolar-v175-5918/&do=findComment&comment=3409498

PR #104, opened 2018-07-03
AirlineKuisine added to Support and Antenna bug fix

PiezPiedPy:
* Make AirlineKuisine work with CryoTanks and USI support.
* Make CryoTanks and USI support work together.
* Fixed antennas bug having no science data rate in languages other than English


PiezPiedPy:
See PR 3 on AirlineKuisine https://github.com/JadeOfMaar/AirlineKuisine/pull/3

PR #115, opened 2018-07-08
Same process in multiple slots

PiezPiedPy:
* Configurable parts can now contain the same process in multiple slots.
* Fixed MRE not running when shielding is full or does not exist on a vessel.
* MRE process now outputs a small amount of CO2.


PiezPiedPy:
screenshot0

PR #127, opened 2018-07-10
RemoteTech: vessels without power will no longer relay

MartinDomig:
Vessels without electricity are blacked out in RemoteTech (same meachanism is used with CME storms), that way RT will no longer allow communication links to be routed through a vessel without power.

To do this, I added a variable `powered` to Vessel_info, my guess is that this info might be useful elsewhere too.

Also, moved RemoteTech specific stuff (variable names like "IsRTActive", "IsRTBroken") into RemoteTech.cs to avoid RemoteTech specifics to be scattered through the entire code base.


PiezPiedPy:
Could you add a brief description of the fix to the CHANGELOG.md file 😉

MartinDomig:
Cool, first time I added to a PR after it was created 👍

MartinDomig:
monitor
And this is what it actually looks like in practice

PR #180, opened 2018-08-27
Tutorial

MartinDomig:
Added a short description that should get a new player through their first mun mission with kerbalism.

Someone proof-read it, please.


PiezPiedPy:
Looks good to me 😉

docs/tutorial.rst
```diff
@@ -0,0 +1,50 @@
...
+
+In order to survive with Kerbalism, Kerbals need enough food, water and oxygen for the duration of the entire mission. Command pods contain enough of these to last for 5 days, which is just enough for a quick trip to Mun, like a flyby or a touch and go landing. Anything beyond that will require additional supplies, so stick to Mun - and Mun only. Do not stay for more than a few hours. And don't even think about Minmus, not yet. Any deviation from that will get your crew killed for a number of reasons, starvation just being one of them.
+
+The most important resource on any crewed vessel, after oxygen, is electricity. Without electricity, your crew will suffocate, freeze or burn to death, whichever kills them first. Not having any food left won't be your number one problem when climatization fails and you're about to burn up in the sun. So, let's take care of that first.
```
SylvainLasnier: I had some problem without radiator: condensation until electrocution

PiezPiedPy: @SylvainLasnier You need the Humidity controller not radiators

MartinDomig: That is odd. For condensation you have to run an ECLSS with a humidity controller.

While it is true that a radiator would reduce the temperature, cold air actually can store LESS water, so cooling air down would actually INCREASE condensation. However, I don't think that kerbalism simulates that.

SylvainLasnier: Understood. Thanks

PiezPiedPy: Your correct, Kerbalisms humidity process does not take into account the habitat temperature.

PR #184, opened 2018-08-29
Fixed the EC issue on high speed time warp.

MartinDomig:
Basically, there were 2 problems:

1. The high-speed warping logic didn't kick in fast enough, it could happen
   that a few simulation ticks (~30min each) were simulated as being in
   complete shadow. Vessels without ridiculously overpowered batteries can't
   handle that.

2. When accelerating to high-speed warp WHILE IN SHADOW, the cached value
   of solar_flux was 0. This is bad, as this is a factor for the EC
   production of solar panels, and you know what happens when you multiply
   anything with 0, right? Right.

So, the fix is: when high speed warping is detected, recalculate the
sunlight as a fraction of $TheTimeSpentInShadow / $OrbitalPeriod. This code
was already there, but executed too late (see above). Secondly, re-calculate
solar_flux, without accounting for day/night (this is new).

This fix is "Works for me"-certified (tm) and I haven't lost any kerbals since.

Four Kerbals died during hunting this bug down, may they rest with the kraken.


PiezPiedPy:
This has got to be one of the best PR's ever.

Issue #186, opened 2018-08-30
Add Vessel Type Filters to Vessel List


eberkain:
The vessel list can become quite cluttered, the groups feature is nice but a simpler filter would be greatly appreciated.  I picture it just using the same vessel designations and icons from map view/tracking station.  It would just be a row of icons that you could toggle on and off each one to have those types of vessels show up in the vessel list, so If I wanted to see all my bases, landers and spacecraft I could just toggle those on while leaving off the relays and probes.

MartinDomig:
<img width="476" alt="bildschirmfoto 2018-09-02 um 20 54 12" src="https://user-images.githubusercontent.com/4105874/44959692-dbef1e80-aef2-11e8-93a0-22464937bf24.png">
<img width="467" alt="bildschirmfoto 2018-09-02 um 20 54 50" src="https://user-images.githubusercontent.com/4105874/44959694-de517880-aef2-11e8-9594-b653258ad62c.png">

eberkain:
Ohh that looks nice

lordcirth:
As it has been merged, should this issue be closed?  Also nice work!

PR #187, opened 2018-08-30
Search / Filter improvements for the Monitor

MartinDomig:
The Monitor allows to search for
- Groups (as before)
- Name of the vessel
- Name of central body of the vessel
- Partial words (you don't have to type out minmus all the way to
  the end to find your vessels orbiting it)
- Search is case insensitive now

This could also be a first step towards adding vessel type filter buttons (#186). Now that the filter is always visible in the montior, it would be the ideal place to put those icons: make the vessel type icons aligned bottom left, put the search text input next to it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 * Stock antennas can now be controlled by automation (Yaar Podshipnik)
 * Devices shown in the device manager are now sorted (Sir Mortimer)
 * Fixed the EC issue when accelerating to extremely fast time warp while a ship is in shadow (Sir Mortimer)
-* Improved vessel search in monitor, now you can search for the name of the central body and the vessel name, too (Sir Mortimer)
+* Iproved vessel searching in monitor: you can search for the name of the central body and the vessel name (Sir Mortimer)
 
 ### Known Issues
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Stock antennas can now be controlled by automation (Yaar Podshipnik)
 * Devices shown in the device manager are now sorted (Sir Mortimer)
 * Fixed the EC issue when accelerating to extremely fast time warp while a ship is in shadow (Sir Mortimer)
+* Improved vessel search in monitor, now you can search for the name of the central body and the vessel name, too (Sir Mortimer)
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -9,7 +9,7 @@
 * Stock antennas can now be controlled by automation (Yaar Podshipnik)
 * Devices shown in the device manager are now sorted (Sir Mortimer)
 * Fixed the EC issue when accelerating to extremely fast time warp while a ship is in shadow (Sir Mortimer)
-* Iproved vessel searching in monitor: you can search for the name of the central body and the vessel name (Sir Mortimer)
+* Iproved vessel search in monitor: you can search for the name of the central body and the vessel name (Sir Mortimer)
```
PiezPiedPy: Typo 😛 Iproved ? Improved

src/UI/Monitor.cs
```diff
@@ -312,16 +304,55 @@ void Render_menu(Vessel v)
...
+
+		Texture2D getVesselTypeIcon(VesselType type, Boolean disabled = false)
+		{
+			// TODO use proper ship type icons here. There should be a way to fetch those from
+			// KSP, but I don't know how or where to get them
+			switch(type) {
+				// this is what it will look like once someone has found the correct icons...
+				// case VesselType.Base: return disabled ? Icons.base_grey : Icons.base_white;
+
+				case VesselType.Base: return disabled ? Icons.battery_yellow : Icons.battery_white;
```
MartinDomig: I need someone to actually provide the correct icons here, these are just placeholders. However, as I currently a) don't see any icons at all in my own builds (mac...) and b) wouldn't know where to get those icons from (I'm sure they're somewhere readily available within the depths of the KSP/Unity API...) I can't do that myself.

Help, anyone?

MartinDomig: This would fix #186

PiezPiedPy: I'll have a go later tonight if I have some spare time but will more likely be Saturday

CHANGELOG.md
MartinDomig: That's the least of my problems ;)

PiezPiedPy: lol

PR #193, opened 2018-09-07
Better SSPX support:

doktorjet:
1) Balanced SSPX greenhouses to be twice as efficient as Kerbalism greenhouse. See [this post](https://forum.kerbalspaceprogram.com/index.php?/topic/172400-131145-kerbalism-v180/&page=32&tab=comments#comment-3446891) for details.
2) Added missing exercise equipment for PXL-E 'Mercury', PXL-F 'Pilgrim' and PFD-C 'Coriolis'.
3) Fixed modulemanager syntax for sspx-utility-125-1 nerf patch. "@" is for editing existing values. No prefix is for adding new ones.

PR #194, opened 2018-09-08
Not human but 25% as big

PiezPiedPy:
Kerbal's apparently use 1/16 of the resources that a human does, due to their size and day length, since Kerbalisms rates where based on a humans use and earth day of 24Hrs, I have recalculated them to a Kerbal's scale and day of 6Hrs.
Breathing, Drinking, Eating, and associated processes such as the Scrubber and Recycling etc. including the Greenhouse have been modified.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 
 ### Changes since the last release
 
+* Kerbal rates for Breathing, Eating etc where based on a Humans use, they have been recalculated based on a Kerbal using 1/4 of a Humans use (PiezPiedPy)
 * A dump valve has been added to the Fuel Cells (PiezPiedPy)
 * Reliability added to the active shield and slightly increased its effectiveness (PiezPiedPy)
 * Vessels that where missing on the monitor due to old SaveGames with RemoteTech antennas fitted, bug fixed (PiezPiedPy)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 
 ### Changes since the last release
 
-* Kerbal rates for Breathing, Eating etc where based on a Humans use, they have been recalculated based on a Kerbal using 1/4 of a Humans use (PiezPiedPy)
+* Kerbal LS rates have been recalculated based on 1/16 of a Humans consumption due to their size and day length (PiezPiedPy)
 * A dump valve has been added to the Fuel Cells (PiezPiedPy)
 * Reliability added to the active shield and slightly increased its effectiveness (PiezPiedPy)
 * Vessels that where missing on the monitor due to old SaveGames with RemoteTech antennas fitted, bug fixed (PiezPiedPy)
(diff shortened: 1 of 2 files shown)


steamp0rt:
I'm worried about balance considerations.

eberkain:
I thought it was 1/4 based on respective days.  
A kerbal uses 1/4 the life support as a human per day.  And you assume the kerbal days is 6 hours and the human day is 24 hours, then the rates are actually the same.

theJesuit:
That does make sense around 6 hour days... but the volume of a kerbal calc also makes sense.

If consumption is quartered (hung on tables,  drawn in code) then kerbalism's current greenhouse will go from supporting .5 kerbal day over a harvest to 2 kerbals.

PiezPiedPy:
@eberkain If a human uses 10kg over 24hrs and a Kerbal uses the same human rate of 10Kg over 6 hrs then the Kerbal has actually used 4 times more, not less 😉 

@theJesuit The greenhouse was recalculated so that it still supports 100% O2 and 0.5 food requirements.

With this PR the SPRX PR #193 makes more sense since @doktorjet based his calculations on a Kerbal using 1/4 of a humans. It also means the Kerbalism greenhouse also falls into spec.
If we use the current scale of 1 Kerbal = 1 Human then the greenhouses when calculated from real world applications are actually 4 times too small.

Also storage now fits better with resource use and how long for example a container of O2 lasts as the amounts stored where calculated in a previous PR using actual volumes of the containers, pressures  and densities of the resources IRL,

@steamp0rt All I can suggest now is for people to try out the profile and leave their thoughts 🤷‍♂️

MartinDomig:
I have to agree with @PiezPiedPy on this. With the new consumption rates, the resource container sizes are very much comparable to what they used to be in TAC-LS. Also, building vessel that can support a crew outside the Kerbin SOI is a lot more practical with this. It still isn't easy, and it shouldn't be, but it won't require 100+ parts just to keep the crew alive any more.

I suggest to take the current profile, rename it to "Real Scale Life Support" (or something similar), and go with these new configs in the default profile.

PiezPiedPy:
I was thinking of making two profiles but it will take quite a bit of effort to go through all the support configs and add the second profile to them.
Not hard to do, will just take a bit of time. I think I'll make a start this weekend coming.

SirMortimer:
If Kerbals are just 1/4th the size of humans, shouldn't the comfort settings in Settings.cfg be changed, too?

`  // quality-of-life settings
  IdealLivingSpace = 40.0             // ideal living space per-capita in m^3
  ComfortFirmGround = 0.4             // firm-ground comfort factor
`

40 m^3 seems a bit roomy for someone the size of master yoda.

doktorjet:
> If Kerbals are just 1/4th the size of humans, shouldn't the comfort settings in Settings.cfg be changed, too?

No, they are 1/2 size of human. 90 cm in height, while average human is 1.8m.
1/4 rate is because of consumption rate is proportional to ^2, not ^3.

SirMortimer:
In my opinion this is a balanced change. @steamp0rt I vote to merge this (if there is such a thing as a vote)

steamp0rt:
k

PR #195, opened 2018-09-11
Use Community Category Kit if it is installed

SirMortimer:
* All kerbalism containers moved to container category
* Greenhouse, ECLSS, Gravity ring moved to life support
* All other parts appear in their respective stock categories only,
  the Kerbalism category is now empty and disappears.


PiezPiedPy:
Any chance you could add the extra tags for the parts search filter like they are in the System.Misc.cfg file
Also can you use FOR[Kerbalism] rather than FINAL as FINAL should really be reserved for end users ;)

Issue #59, opened 2018-06-15
toggleable setting to default on transmission enabled


Gordon-Dry:
When adding science to the transmission queue I sometimes ask myself, why is it not just set to be transmitted by default?

So I request this as a setting in the settings.cfg:

TransmissionDefault = True

But ofc it would be just a little bit better if this behaviour could be toggled ingame.

Gordon-Dry:
Just for further explanation; some "science management mods" like ScienceAlert pop up the message that a science is available again just after it has been transferred to the Kerbalism queue.
If it would be transmitted immediately (as long as a connection exists) this would not happen "so badly".

But I should mention that I also opened an issue on ScienceAlert because of this, see:
https://github.com/linuxgurugamer/ScienceAlert/issues/4
This one has already been closed, but IMHO it shouldn't been.

SirMortimer:
See #196 - I added a new setting that will affect all new science files. If that option is set, new experiment data should be transmitted as soon as possible. Default is on.

Issue #185, opened 2018-08-30
Desactivate message by default and pause when there is one


SylvainLasnier:
Hi,
Due to Kerbalism rules, most of vessel are unmanned probe and alert message dont matter a lot for them. And with lot of probe, it become confuse.
In another hand alert messages are vital for manned station. But it is easy to miss some critical message when your are wrapping to the next maneuver.

So, is it possible to desactivate message by default in the cfg panel?
Is it possible to pause the game when a message is display? (option like alarm clock)
Nice to have, is it possible switch to the vessel by clicking on this list?

br

eberkain:
I think I'm following you so I can try to answer.  

> So, is it possible to desactivate message by default in the cfg panel?
You can deactivate messages on a per vessel basis, but they are all on by default.  I'm not sure there is alot of value in being able to configure this before hand.  You could have it inherit when a vessel is undocked/decoupled, but that could lead to situations where a message is disabled that you did not intend to have disabled which might kill your crew without letting you know there was a problem.  Its probably safer to always have them on by default and toggle on a per vessel basis in-flight.  

> Is it possible to pause the game when a message is display? (option like alarm clock)
A message will break out of warp currently, but this has a conflict with some auto-warping mods like MechJeb which immediately starts warping again.  This may be something that could be looked at, but probably belongs in a separate issue with clear details. 

> Nice to have, is it possible switch to the vessel by clicking on this list?
That could be a good feature to add to the vessel list, also the ability to target any vessel anywhere off the list would be handy.  Not sure how the UI would be implemented though, that list is already jam packed with icons and text.

PiezPiedPy:
switching vessel by clicking on the list in the panel might be problematic as that is how a vessel is selected for viewing its details and using right click is just as bad because that is used to hide the panel.

An option in the cfg panel to pause the game on a message might be possible.

eberkain:
The option to sort the list A-Z or Z-A or default would be nice.  :)  

Vessel switching and vessel targeting would both take new buttons i think, maybe you could squeeze them onto the bottom tab list with info, data, auto, etc...

SirMortimer:
> switching vessel by clicking on the list in the panel might be problematic

mabe do it with a button in the vessel details?

SylvainLasnier:
It make sense

SylvainLasnier:
> I think I'm following you so I can try to answer.
> 
> > So, is it possible to desactivate message by default in the cfg panel?
> 
[...]

PR #196, opened 2018-09-13
Moved a bunch of Settings from the config file into the GUI

SirMortimer:
Screenshots say most.

This isn't thoroughly tested yet, however, I'm fairly confident. Things to test, specifically:

* Radiation levels, are they what they should be
* CME Storms (days between storms, storm durations)

The other settings should be working OK, as far as I can tell.

This PR will also fix part of #185 (the message notification presets are new, that config didn't exist yet)
Kerbalism
Kerbalism (Advanced)

Also, I'm not sure what to do with these settings, heck I don't even know what they're supposed to do after looking at the code:

`
  // pressure settings
  PressureFactor = 10.0               // pressurized modifier value for vessels below the threshold
  PressureThreshold = 0.9             // level of atmosphere resource that determine pressurized status

  // poisoning settings
  PoisoningFactor = 0.0               // poisoning modifier value for vessels below the threshold
  PoisoningThreshold = 0.02           // level of waste atmosphere resource that determine co2 poisoning status
`

@steamp0rt don't merge just yet, this needs to be tested a bit (which is why I created this PR).


PiezPiedPy:
Brilliant, saved me a job 😄 
btw Preferences.cs  should be in the System folder, the Modules folder is for PartModules 😉

src/Modules/Preferences.cs
```diff
@@ -0,0 +1,315 @@
...
+
+		public override string Section { get { return "Kerbalism"; } }
+
+		public override int SectionOrder { get { return 0; } }
+
+		public override string Title { get { return "Common"; } }
+
+		private static PreferencesBasic instance;
+
+		public static PreferencesBasic Instance
```
PiezPiedPy: maybe rename this to fetch to fit in better with KSP coding style, no idea why squad thought it a good idea to do such a thing 🤷‍♂️

PiezPiedPy: same with the other occurrences of Instance

SylvainLasnier:
Awesome!

SirMortimer:
> Brilliant, saved me a job 😄
> btw Preferences.cs should be in the System folder, the Modules folder is for PartModules 😉

yeah I know... didn't bother moving it :P

src/Modules/Preferences.cs
SirMortimer: Actually I stole that bit from kOS. No idea why they named it that way... Instance refers to the singleton pattern I use a lot in different languages.
I'm new to C#, so bear with me...

SirMortimer:
This should also fix #59.

screenshot4

src/Modules/Preferences.cs
PiezPiedPy: Instance should also be used also in C#, its just that Squad uses fetch for some unknown reason

SirMortimer:
Disabling of stock science seems to have no effect. Other than that, I think it's OK

steamp0rt:
Fix the stock science, I guess?

steamp0rt:
As soon as you do, i'll merge this and release 1.9.

Issue #198, opened 2018-09-13
Use Courage, Stupidity and Veteran crew status


SylvainLasnier:
Could we use courage, stupidity, level and veteran crew status to alter the gameplay?
For example:   
* a stupid engineer is less able to repair a failure
* a stupid scientist could loose data to analyse in Lab
* a stupid pilot could fail to maneuver ship
* a brave crew accept more radiation, poisoning and other kerbalism miseries...
* a veteran could detect some issue without sensor
...
TBD

SirMortimer:
I'm currently tinkering with using courage+stupidity to affect process rates on an individual basis. The effect would be that two kerbals on the same vessel would each use slightly different amounts of food, water and oxygen, for instance. Note that Kerbalism already has things that depend on individual Kerbals, like the time it takes to die from starvation or from radiation poisoning.

I wouldn't know if having stupid scientists loose data in the lab wouldn't take it too far, but repairs already require Engineers with a minimal level, depending on the component that needs to be repaired. 

What will be worth looking into is having the crew levels affect the efficiency of a lab or a drill, like stock KSP does it. A level 5 scientist processes data quicker, a level 5 engineer helps a harvest drill to operate more efficiently.

HaullyGames:
I think is a good idea.
The KerbalBreakdown already create a random loose data event when vessel has data.
Also, why don't use stupidity to affect the KerbalBreakdown system? using it as a multiple to occur KerbalBreakdown.wrong_valve in Breakdown method?
The Misc.Breakdown method receive ProtoCrewMember c, but it is doing nothing with that.

SirMortimer:
I'm intrigued :)

How about this: let's assume kerbals aren't perfect, and can make mistakes when under stress. Stress is almost zero early in a mission, but as they stay in space for a very long time, their stress level rise - as does the likelihood for them making a mistake (a breakdown). That likelihood can be affected by a kerbals individual stupidity, more stupid -> more errors. Now, if the Kerbal really DOES make a mistake, that in turn raises his stress level...

This, in combination with a setting to turn off such stress induced breakdowns and maybe a slider to  tweak the probability of an error happening once per week or so.

SirMortimer:
I guess when #203 and #204 are merged, this can be closed.

SirMortimer:
@steamp0rt this can be closed.

PR #203, opened 2018-09-17
Make use of a Kerbals individual personality.

SirMortimer:
Some processes already used a random element in degeneration (radiation posoning, humidity, eating) which was based on the first name of a kerbal. This PR now uses the personality traits (courage and stupidity) instead, and adds a new property to a rule: individuality. The idea stems from #198.

Individuality is a value that is supposed to tweak the rate of a process for individual kerbals by just a bit. The default profile adds 1% of individuality to the usage of food, water and o2, 15% to radiation poisoning and 10% to stress compensation. The individuality for a kerbal won't be the same for two different rules, so a crew that eats less might drink more instead.

In addition, being under high levels of stress no longer is without consequences. Errors happen, and they happen in the Kerbal universe too: depending on the stress and stupidity of a kerbal, there now is the possibility for a breakdown even in relatively low stress situations. Breakdowns can happen to anyone, at any time - however the probability of them happening depends on the stress they're in.

Stress related breakdowns can be configured in the save game and are off by default.

screenshot1


steamp0rt:
Nice! Good job.

PR #204, opened 2018-09-18
Laboratory / Harvester efficiency increases with crew level

SirMortimer:
The laboratory will now give a 10% bonus in processing speed for each level that any cew member in it that is above the required minimum. If you have a lab that requires a level 1 scientist, having a level 5 scientist in it will give you a 40% bonus. The effect accumulates if you have more than one crew above the required minimum. The maximal achievable bonus is clamped to 70%.

Same goes for the harvesters, but here you need engineers on the vessel, not in the part - for obvous reasons. Just having any engineer on board will give you a 20% bonus, with additional 20% for each level. So a level 1 engineer will yield 40%. Again, the maximum bonus is clamped to 70%. This should be an incentive to have mining stations crewed with engineers, to give them more purpose in the game.


src/Utility/CrewSpecs.cs
```diff
@@ -63,6 +63,34 @@ public bool Check(ProtoCrewMember c)
 			return trait.Length == 0 || (c.trait == trait && c.experienceLevel >= level);
 		}
 
+		// return sum of al bonuses for crew on vessel
```
PiezPiedPy: How about using ///<summary> instead of a //comment 😉

SirMortimer: Nice, didn't know that one. One has to apprechiate how MS manages to reinvent wheels all the time. Many other languages use /** ... */ for that...

PiezPiedPy: Uses Intelisense, handy because you can hover over the class, method etc elsewhere in the code and see the comment, I now use it on nearly everything 😃

SirMortimer: Yeah, but this is the poor mans version of what other IDEs offer, and you have to write even more control gibberish to make it work.

PiezPiedPy: different languages comment in different ways, better than nothing though 🤷‍♂️

PR #209, opened 2018-09-20
Advanced Medicine

SirMortimer:
The amount of radiation one person can endure during their entire life span is limited. In fact, astronauts on a mission to Mars would be exposed to at least 60% of the total radiation dose limit recommended for their career during the journey itself. Recovering a Kerbal on Kerbin will no longer magically cure all radiation.

In addition, this fixes the DeepFreeze cheat that lets you reset a Kerbals radiation status by freezing and thawing them. When frozen, Kerbals won't accumulate any radiation, but they will no longer be cured when they thaw.


theJesuit:
Should this have been called anti-medicene :)

lordcirth:
This is an option that is off by default, right?

theJesuit:
Ah and as I understand branches this makes sense.

Will there ever be a cure?

SirMortimer:
First things first:

This is hardcore. I know. That's why it is purely optional, can be configured in the settings screen and is off by default. It also is why I was hesitating even creating this PR. Here are some pros and cons:

* This might very well be too hard for career games, where you'll end up literally burning through your crew to the point that the radiation alone of an additional orbit might kill them. If you can't rescue new crew from orbit, you'll have to spend loads of credits recruiting new crew and upgrading the astronaut complex, and we all know how much extra money there is available in a career game to spend for such luxuries: none.
* There is no way to manage the health status of your Kerbals back at KSC. If you don't remember which Kerbal can endure an additional mission, you won't find out until your rocket sits on the launch pad with the kerbal in it.

On the other hand:

* Very recently, Kerbalism has been made a bit easier. You don't need to bring as much resources as you had before, you need less greenhouses and they produce food practically right after launch.
* This fixes a cheat that currently is possible with DeepFreeze pods, which effectively are regeneration chambers: when you freeze a suffocating kerbal that is close to radiation poisoning, at the brink of starvation and at maximum stress level, you'll get it back brand new when you thaw it.
* This opens new possibilities, like adding sickbay part modules that cure radiation poisoning, possibly requiring a rare resource that has to be harvested. Maybe put it into the hitchhiker and in the gravity ring, once a sufficiently high tech has been researched.
* It adds a real incentive to actually add shielding to your stations (usually mine don't have it if they're close enough to kerbin to allow a crew return before a CME hits) 

And there is yet another idea that I like, I'll just paste this from discord:

> What would you say if radiation would fade gradually over the time? Additionally it could take increasingly more time, the higher the dose. Like 1 year for 50%, but for 75% it would be 2 years, for 87,5% 3 years and so on.
> And/or the radiation wouldn't ever go back to 0%, but for example if you came back with 50% it would go down to not less than 10%, and if you would come back with 90% it wouldn't go below 50% anymore.

What do you think?

theJesuit:
Not sure about the fading personally as a quick thoughg it may be possible to game the system to have it fade at the same rate as it is being added even sitting on that 10%.  I'm just think of the IRL consequences that kerbalism is trying to game. 

I'm still not sure why people worry about retiring kerbals?  An astronaut IRL who travels to Mars is unlikely to be selected for the mission to Ceres or even Saturn in the NASA sense.  That MAY change with space colonies and privatisation.  Is there an expected life span of kerbals? Should there be?

PiezPiedPy:
Sickbay idea is sick 🤢

steamp0rt:
What will the sickbay do?

SirMortimer:
Technically speaking: it will run a process that provides an arbitrary offset to a rule, affecting only the kerbal currently in that sick bay.

In plain english: it will use resources to cure radiation poisoning.

Currently my idea is to make a Radiation Detoxication Unit (RDU) which uses loads of EC and oxygen to provide, to a degree, relief from radiation poisoning, output will be waste water (to preserve mass). The bay shouldn't be able to cure 100%, and maybe slow the healing process down at a treshold. The same part, with different processes, could also be used to heal other things. I could imagine maybe a Holo Deck that uses one nuclear reactor worth of EC to slowly reduce the stress level of a kerbal.

It would also be possible to have a RDU use a rare material that needs to be harvested. Maybe it would require RadX that only exists in the clouds of Jool or in Eve's explodium sea... we can (and maybe should) be very mean with this.

I'm not planning to make a healing process cheap, if the input materials aren't rare it will need loads of them. Also, the sickbays will require a quite high tech level.

It's not done yet, and the current implementation isn't going to work as expected, which I just found out recently.

theJesuit:
Like the radiation detox idea.  Just a thought based on the current 26 Sep blurb that O2 in and waste H20 doesn't seem balanced.

SirMortimer:
If you're referring to the current numbers in the process, don't take them seriously. They're made up numbers. Right now I'm working on just getting a process to run and provide the cure only to that kerbal.

theJesuit:
Wasn't numbers but simly wondering where the hydrogen could come from - unless radiation poisoning induces nuclear fissioning of oxygen to hydrogen.

SirMortimer:
Anyone got an idea on how to setup the healing process? Currently I've just put in some random guesswork, using O2 and Electricity, producing CO2 and The Cure™:

```
Process
   {
     name = radiation detox unit
     modifier = _SickbayRDU
     input = ElectricCharge@5
     input = Oxygen@0.2
     output = WasteAtmosphere@0.2
     cures = radiation@0.01
   }
```

This is just made up on the spot without any thought. When finished, the process should fit into a narrative that goes like _"There was a small incident at R&D involving a nuclear reactor and a ruptured tank. While experimenting with hydrogen and an experimental aluminium alloy, Dustin Kerman tripped over a cable, resulting in a power outage that made international news. What didn't make the news was the green glow emitting from Dustins skin. Several reenactments of that experiment suggest that, when applied in zero gravity, this reaction could regenerate body cells damaged by radiation.

Warranty void if used under the influence of gravity."_

PiezPiedPy:
Regarding mass and densities with an input of 0.2 Oxygen and assuming the Carbon is from the Kerbal's dead cells you should have an output of 0.1987945992 WasteAtmosphere.

PiezPiedPy:
Also maybe the sickbay should have a built in scrubber and have an output of CarbonDioxide rather than WasteAtmosphere as that ends up putting extra strain on the scrubbers.

PiezPiedPy:
I might of been a bit tight with the amount of radiation the sick bay cures, at 0.001 rad/h a Kerbal will never be cured if the vessel has a nuke engine as it puts out 0.02 rad/h, so maybe the rads cured should be upped to 0.1 rad/h = radiation@0.000027777777.
At that increased level, a day on a vessel with one nuke engine would expose a Kerbal to 0.12 rads and thus the Kerbal would need 1.2 hrs in the sick bay to cure the exposure. Or to put it another way 5 days exposure would need a day in sick bay.
Maybe an even higher number would be better like 0.25 rad/h cutting down the time in sick bay to half an hr 🤷‍♂️

docs/modders/profile.rst
```diff
@@ -141,6 +141,10 @@ Rule and Process rates can be influenced by the environment, habitat conditions
 +-----------------+--------------------------------------------------------------------------------------------------------------+
 | per_capita      | the inverse of number of crew members, the effect on rates is a division by number of crew members           |
 +-----------------+--------------------------------------------------------------------------------------------------------------+
+| zerog           | 1 if the ship is above the atmosphere of a body, 0 if in atmospheric flight or landed                        |
```
PiezPiedPy: Me like 😃

PiezPiedPy:
files have moved to src/Kerbalism/ folder ;)

steamp0rt:
Fixed conflicts for you.

steamp0rt:
Tell me if/when this is ready

SirMortimer:
Wow thanks, I was just going to merge master and couldn't find out why everything is OK now :)

SirMortimer:
> Tell me if/when this is ready

Will do. Still in testing.

SirMortimer:
accidentally. it was late...

SirMortimer:
Done.

We might need to adjust a few configs a bit, but I'd wait for feedback.

Merge, please.

steamp0rt:
rebuild it

steamp0rt:
actually i can do it nvm

SirMortimer:
> DAMMIT YOU FORGOT A CHANGELOG

yip.

PR #216, opened 2018-10-02
Habitat changes, equalize changes and other small changes.

HaullyGames:
Changes list:
Resource
- Show all resources in debug mode

GravityRing
- Check status to not create multiple calls.

ResourceBalance class
- The class is hosted by the file Resource.cs (it won't affect other class into Resource.cs)
- This class will be responsible for Equalize resource into a vessel.
  - Those resources are defined by a array in code, simple to be add new resources to be balanced.
- The class do a Equalize\Venting of all resources share the same Math, with different logic.
- The ResourceBalance has 99.9999% precision in the balance process (100% is heaviest, running to many times for fractions of amount)
  - With 99.9999% balance, it produce or consume some amount of resource,
    - This amount is unnoticed by the player since it is like 0.0000# of amount

Habitat
- Fix Nitrogen missing when Kerbalism is added in existing game
- Review and fix small triggers for ResourceBalance
- Control IVA (Spawn\Destroying) based if habitat is habitable or not (not inflatable habitat will always be able to see IVA)
  - IVA in KSP also control the small window in the right down side of the screen, this is a standard behavior
  - Inflatable hab with less than 100% and with crew won't show the IVA, but it will stay passable with CLS
    - Maybe you want to move the crew for a better habitat😝
    - This inflatable behavior + IVA KSP behavior, you won't see the kerbal face in the right down window when it is not 100%
- Optimize performance in FixedUpdate(): less method call per frame
- Equalize\Venting logic has been moved from Habitat to a separated class that handle Equalize and Venting for whole vessel
- Equalize was renamed to Pressurize, as this would only check pressurizing
- Venting was renamed for Depressurize, since this would only verify the depressurizing
- Has 3 ways to inform ResourceBalance if this habitat needs to be equalized
  - Status has been changed (enabled to depressurizing or any other)
  - Crew has been transferred for the habitat (this automatic Turn On the Equalize, Bring more resource if necessary and available)
  - Enabled habitat with crew is losing pressure (this automatic Turn On the Equalize, Bring more resource if necessary and available)
- Shows perct deploy\inflated

Air Pump
- New Process: Because pressure control doesn't work in breathable environment, this process will only pump air into the vessel.


src/Resource.cs
```diff
@@ -26,6 +23,10 @@ public Resource_info(Vessel v, string res_name)
 							amount += r.amount;
 							capacity += r.maxAmount;
 						}
+#if DEBUG
```
PiezPiedPy: Could you give this it's own Directive 😉 something like RES_DEBUG

src/Resource.cs
```diff
@@ -284,9 +279,7 @@ public Resource_recipe()
 			this.left = 1.0;
 		}
 
-		/// <summary>
-		/// add an input to the recipe
-		/// </summary>
+		// add an input to the recipe
```
PiezPiedPy: Why remove the summaries ? summaries are preferred over comments

src/Resource.cs
```diff
@@ -550,6 +536,242 @@ public static void Transform(Vessel v, Resource_recipe recipe)
 		static Dictionary<Guid, Vessel_resources> entries;
 	}
 
+	// equalize/vent a vessel
```
PiezPiedPy: what do you think about putting this in its own file for readability 🤷‍♂️

PiezPiedPy:
Could you add a line to the changelog.md with a brief description of your changes

src/Resource.cs
HaullyGames: I added the code in Resource.cs because this class hosts all activity related to the resource (Cache, Info, Recipe), I thought it was the right place to add a class that would control resource balance.

src/Resource.cs
HaullyGames: Adding back.

src/Resource.cs
HaullyGames: Sorry, but I'm not sure how I do it, I can search to be sure, but if you can do this or send me the instructions, I'll do it.

src/Resource.cs
PiezPiedPy: it was just a thought, usually better to keep large classes in their own file, not a big deal tbh

src/Resource.cs
```diff
@@ -288,7 +290,9 @@ public void Input(string resource_name, double quantity)
 			}
 		}
 
-		// add a combined input to the recipe
+		/// <summary>
+		/// add a combined input to the recipe
+		/// /// </summary>
```
PiezPiedPy: typo

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@
 * Harvesters will work better with an engineer on board. (Sir Mortimer)
 * Fixed another icons sometimes not displaying bug (PiezPiedPy) 
 * SSPX 2.5m Greenhouse now producing food at the expected rate. (theJesuit)
+* ResourceBalance run the pressurizing\depressurizing (old "equalize\venting"), it gives priority to habitats with crew(HaullyGames)
```
PiezPiedPy: Could you word this for users to understand what's changed, the changelog is mainly for the players not for us coders 😉 hope I'm not being a pain.

Something like ' Habitats equalize/venting function changed to pressurizing/depressurizing, crewed habitats have priority while multiple habitats are pressurizing. New AirPump process added blah blah etc.

steamp0rt:
Fix conflicts please! (Should be easy, i just moved files to src/Kerbalism)

steamp0rt:
Conflicts fixed

steamp0rt:
Tell me if/when this is ready

steamp0rt:
capture

PR #213, opened 2018-09-23
Refresh CommNet status of unloaded vessels

leomike:
Code based off @HaullyGames's work. Resolves #212.

PR #223, opened 2018-10-07
Solve potential null references

leomike:
Fix potential null references introduced by #213. Didn't update the changelog given this change had not been released yet.


Gordon-Dry:
Would be cool.

PR #250, opened 2018-10-23
Because there are more than 2 types of Kerbals (as parts)

Gordon-Dry:
Making History!


theJesuit:
So male kerbals are a different species... go figure.

PR #251, opened 2018-10-23
Kerbal eva rt antenna (active)

Gordon-Dry:
Now a Kerbal on EVA can send the EVA report to KSC by radio ...

Works with CommNet and RemoteTech ...

Antenna range is ~5km - so the vessel / rover / base would be the relay then ...

https://cdn.discordapp.com/attachments/418222516386004994/504149562991181825/screenshot_2018-10-23--06-28-52.png
https://cdn.discordapp.com/attachments/418222516386004994/504149640619491338/screenshot_2018-10-23--06-29-29.png


Gordon-Dry:
On hold, needs a further test.

PiezPiedPy:
nice update to the Remote Control Radio, don't forget changelog 😉

GameData/Kerbalism/System/Signal.cfg
```diff
@@ -50,11 +50,11 @@
 
 @PART[kerbalEVA*]:FOR[Kerbalism]
 {
-	MODULE
-	{
-		name = ModuleCommand
-		minimumCrew = 0
-	}
+	// MODULE
```
PiezPiedPy: I'm pretty sure this is needed for the remote control of other vessels, as was intended for the EVA remote control radio so the Kerbal's have the ability to remote control vessels when EVA.
Make sure you test this with CommNet 😉

PiezPiedPy:
Maybe the 500m range is a bit lame, a phone can transmit approx. 35km with only mW of power, at 4G network speed

GameData/Kerbalism/System/Signal.cfg
```diff
@@ -62,7 +62,7 @@
 		antennaType = DIRECT
 		packetInterval = 0.2
 		packetSize = 1.35
-		packetResourceCost = 0.5
+		packetResourceCost = 0.375
```
PiezPiedPy: Would be better to set the EC value to the equivalent power usage of the type of antenna that is used. And change the amount of EC given to a Kerbal when EVA to balance with the O2 use, this can be done in the default profile

Gordon-Dry:
I tested it with @HaullyGames Kerbalism15.bin and RT.dll and it worked.
That was the dll from https://github.com/RemoteTechnologiesGroup/RemoteTech/pull/771
I did not get a dll of https://github.com/RemoteTechnologiesGroup/RemoteTech/pull/772 yet.

Gordon-Dry:
Then this also should wait for RT release and #253

HaullyGames:
Hi @steamp0rt and @Gordon-Dry.
I think we are good to merge this PR, am I right?

Gordon-Dry:
Yep

PR #266, opened 2018-10-30
RT - Added API.GetName to support Connection Panel.

HaullyGames:


HaullyGames:
Hi @steamp0rt, This is ready to merge.

PR #283, opened 2018-11-05
Just make it equal to the patching of stock Commnet

Gordon-Dry:
I always wondered what is going on.

Why is not the same kind of patching routine done with the RT antennas as like as the Commnet antennas?

### DIFF
--- a/GameData/Kerbalism/Support/RemoteTech.cfg
+++ b/GameData/Kerbalism/Support/RemoteTech.cfg
@@ -290,9 +290,41 @@
 
 @PART[*]:HAS[@MODULE[ModuleRTAntenna]]:NEEDS[RemoteTech]:AFTER[RemoteTech]
 {
+	@MODULE[ModuleRTAntenna]:HAS[~DishAngle[],#Mode1OmniRange[>0]] // omni
+	{
+		@TRANSMITTER
+		{
+			@PacketResourceCost /= 1000
+		}
+	}
+}
+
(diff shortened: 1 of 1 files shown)


HaullyGames:
Is it right? Have you tested it?

Since we don't use RT EC in Kerbalism (RT doing it by they self), this change won't affect Kerbalism.
Also in RT + Kerbalism, We don't use *PacketResourceCost* in our code! this mean it will be used only if you transmitting through KSP + RT (no Kerbalism).

Maybe this change should be sent to RT team.

NOTE: Looks like I forgot to add to get EC for unloaded vessels with RT, I will add it back, but loaded vessel we didn't consume energy through Kerbalism.

Commit: https://github.com/steamp0rt/Kerbalism/commit/ed29a952b7176d8c215a747f6a0c5453eb56934c

It was
``// consume ec for external transmitters (don't consume for RemoteTech when loaded)	
			if (!(RemoteTech.Enabled && v.loaded)) ec.Consume(vi.connection.external_cost * elapsed_s);``

And to fix the no consumption for unloaded vessel.
Needs to add
``ec += pm.resHandler.inputResources.Find(r => r.name == "ElectricCharge").rate;``
into
https://github.com/steamp0rt/Kerbalism/blob/master/src/Kerbalism/Comms/AntennaInfoRT.cs

Gordon-Dry:
Hmm. I guess you're right.

Actually I'm trying to find what's going on - the NFE update somehow borked my save ...

I'm back on your latest build actually.

But now I have absurd high EC consumption on my station - gotta abandon it and try to bring the last 2 Kerbals home ...

PR #284, opened 2018-11-05
Don't show sickbay modules in automation if module has no capacity

SirMortimer:


madman2003:
@SirMortimer why doesn't the sickbay have a normal on/off button?

Gordon-Dry:
Yeah, the TV running all the time. In every bigger habitat.
And just-in-case the anti-radiation units are also online all the time ...

It's cheaper to check for the stress/radiation levels of the crew members who enter the habitat.
But less sophisticated.

And a button would be - retro. But helpful.

steamp0rt:
Fix conflicts and i'll merge

MartinDomig:
There should be an off option - worked for me. If it isn't there, that's a bug.

madman2003:
@MartinDomig confusing bit is that the sickbay module is behind a part upgrade (which is off in sandbox by default)

madman2003:
@SirMortimer have you abandoned this?

steamp0rt:
F

PR #289, opened 2018-11-09
fix a bracket, fix intends, add a Sickbay_RDU to Duna Direct KRV

Gordon-Dry:
It started with finding a missing bracket in Support/NearFuture.cfg ... at the end

Sickbay_RDU is added to Mars Direct aka. Duna Direct.
Un-consistent intends consistentified.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,11 @@
 
 ### Changes since the last release
 
-* Added sickbay RDU to three additional parts (with 1 slot each):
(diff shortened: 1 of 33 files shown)


madman2003:
Next time can you make the re-indenting a seperate commit for easier reviewing?

Gordon-Dry:
Yes

PiezPiedPy:
that looked like fun :hurtrealbad:

PR #296, opened 2018-11-18
don't fill habitat with waste products when enabling a habitat in VAB

madman2003:
I'd like my kerbals to not die of CO2 poisoning


steamp0rt:
this ok?

madman2003:
@steamp0rt I'm ok with it obviously, but a second opinion on the modifier naming might be a good idea ;-)

PiezPiedPy:
personally I prefer breathable, artificial atmosphere dose not make sense to me, the idea of the breathable modifier is to act as a switch when a vessel is in a breathable atmosphere hence its naming

madman2003:
@PiezPiedPy tell me how to name the inverse and still have it make sense ;-)

madman2003:
an airpump works in a breathable atmosphere, but shouldn't have the breathable modifier
that is the dilemma

PiezPiedPy:
non_breathable 🤷‍♂️

madman2003:
@PiezPiedPy that's possible, but that means swapping the modifiers around, i.e. airpump gets breathable and the rest gets non_breathable

Btw, natural and artificial atmosphere is reasoning from the perspective of what the kerbals will be breathing

PiezPiedPy:
the breathable modifier is for the outside atmosphere not internal vessel atmosphere. 😉

madman2003:
@PiezPiedPy yes, but it was the only inverse modifier, meaning that if the atmosphere is not breathable the process is active, that's confusing especially if you need the opposite modifier as well

madman2003:
And in space there is no atmosphere, so only the vessel internal atmosphere is actually defined

PiezPiedPy:
you have a point there, I would suggest breathable be renamed to non_breathable and add it's inverse breathable.

I think using the name breathable would be better because it is named the same as its variable in code `env.breathable`

PiezPiedPy:
@HaullyGames any ideas for the breathable modifier ?

PiezPiedPy:
@madman2003 another point with using natural or artificial naming is that it does not describe if the external atmosphere is breathable or not, i.e if on eve for example the external atmosphere is non breathable. Using natural or artificial naming aims more at describing what the Kerbal's are breathing in the vessel

some way of naming the modifier to reflect that it is describing the external atmosphere might be better.

PiezPiedPy:
I suggest maybe env_breathable and env_nonbreathable

madman2003:
I wouldn't use abbreviations in a modifier, but that is just me
Bed time for me, i put the breathable/non_breathable version in an another branch/PR

PiezPiedPy:
bed for me too, catch ya later ;)

PR #304, opened 2019-01-02
Fix some bugs and bump version to v2.2.0 for KSP 1.6.0

PiezPiedPy:
also bumped docs and Travis etc. and fixed a couple of bugs.
* Fix Kerbalism parts search filters and missing tab in the VAB/SPH.
* Fix processes not calculating capacities correctly.
* Processes default to a switched on state when added in the VAB/SPH.


steamp0rt:
wow it's 2019

PiezPiedPy:
@steamp0rt Yep 🥇 😛

PR #322, opened 2019-02-24
 Update habitats for the new V2 and DLC pods and fix atmosphere leakage process

PiezPiedPy:
* Added habitats for the new V2 and DLC pods.
* Recalculated habitat atmosphere leakage, was originally calculated for a Human day which is 4x longer.


theJesuit:
Why did we not consider this earlier.

Mainly because it makes it easier?

PiezPiedPy:
I've only just noticed since it was pointed out on a YT video it seemed wrong 🤷‍♂️

PR #323, opened 2019-02-24
Optimized Planner: Part 1 - Chill the stuttering in VAB/SPH

PiezPiedPy:
* Made the planner class static.
* Now updates the sim via events rather than every frame.
* Added an EVA's available indicator to the Planner and Monitor.


PiezPiedPy:
@steamp0rt This can be merged now

PR #334, opened 2019-03-14
Early co2 warning

SirMortimer:
CO2 poisoning happens so fast that you might be in an unrecoverable situation even before you hit the danger threshold, especially on high time warp. Once the co2 level is above 2.5ish% scrubbers won't scrub the co2 in time to stop you from dying. This little fix will give you some more time to react before it is too late.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -359,6 +359,8 @@ sysinfo.txt
 
 Kerbalism*.zip
 
+MiniAVC.xml
+
 # KSP DLLs
 KSPAssets.dll
 Assembly-*.dll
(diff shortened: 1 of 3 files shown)


steamp0rt:
tbh i think there should be an emergency option to flush the atmosphere and re-cycle new air in.

theJesuit:
Yes.  This should sort out humdity too.

SirMortimer:
I am looking at it, but the habitat/atmosphere code is a bit tight to get your head around. I'm optimistic, but no promises.

PR #335, opened 2019-03-15
Preemptive maintenance: components can be serviced

SirMortimer:
Kerbals now can perform preemptive maintenance to avert part failures. First, inspect a component. If it is found to have issues, you can service it. Once serviced, it will be as good as new.
inspection result
servicing


steamp0rt:
nice

Issue #329, opened 2019-03-04
Fuel Cell only works manually


mekaerwin:
When I set the fuel cell to come on at 20% or off at 80% clicking on it says running or off appropriately but it is not actually changing. Manually switching it works. Turning it on or off in device manager under devices also does not work.
I simply take a command pod and add hydrogen tank to side and fuel cell. Launch mission and set up as above. I have no mods other than Kerbalism and Community Resource pack and ran without the expansion installed. I am using ModManager 4.0.2. I have tested on both launch pad and in flight.

I can submit a save but it is simply sandbox start.
I'm not sure if there is anything else you would like.

-Edit
I'm running the latest version 1.6.1

SirMortimer:
can confirm. working on a fix.

PR #336, opened 2019-03-16
Fixed emitters in automation

SirMortimer:
The automation UI listed fixed emitters, like the NERV and lied about being able to turn it on or off. Worse than that, when you try to turn it on or off, the Kerbalism UI effectively renders non responsive.

Fixed emitters (they already have the toggle boolean in the config) are no longer listed in automation, and it is no longer possible to "control" (ha ha) them from there.


steamp0rt:
Why'd you close it

SirMortimer:
because it doesn't fix it. I'll create a new one.
And while I'm at it I'll look at #329, could be related.

Just found out that there are many calls to GameEvents.onEditorShipModified.Fire(EditorLogic.fetch.ship) that crash in flight, have to fix that everywhere. The crashing might be new in KSP 1.6.

PR #341, opened 2019-03-18
Small atmo leak buff

SirMortimer:
I think the atmosphere leaked deserves yet another small buff. I think the surface area estimation of the ISS in the dfeault profile was off by about a factor of 2, if we assume that the ISS isn't some sort of spheroid.


PiezPiedPy:
I took the surface area from Nasa not Raiz estimate. tbh this needs more research.

SirMortimer:
Yep. But it's difficult to get an official number. However, I just looked up some of the ISS modules. They all have diameters from 4m to 4,5m, lengths from 4,21m to 11m. I disregarded the cupola and the Russian modules with their odd shapes and very narrow (<3m diameter) designs. 

Taking the bare sizes, not accounting for the fact that the hull is not made of paper but has some thickness to it (the effective volume is smaller than my result), ignoring that probably not all of the modules volume is actually pressurized, and allowing for a number of "docking ports" that connect the modules and thus are not part of the hull surface, I get a total volume of 850m^3 and a surface of 966m^2.

I really wonder where the 573 m^3 come from, that's only ~20% more than the surface of a sphere with the given volume. It just doesn't check out.
numbers

PiezPiedPy:
I think the documents I used where from 2004, there have been a number of modules added since then so that's probably where the error is, I'm happy to go with the numbers on your chart 😉

SirMortimer:
I pimped that spreadsheet a bit and updated the result. Doesn't change too much, plus it will make future recalculations easier.

PR #344, opened 2019-03-19
Plant bonus

SirMortimer:
There was a variable related to the new plant_bonus missing in the preferences class which caused a compilation error.

Also, MM complains about an issue in Sickbay.cfg, will look at that later.


SirMortimer:
WTF where are all those commit coming from

PR #345, opened 2019-03-19
Plant bonus fix

SirMortimer:
Missing variable in settings class, did not compile.

Also MM complains about an error in Sickbay.cfg, will look at it another time.

Github, please behave nicely this time.


SirMortimer:
Attaboy.

Issue #313, opened 2019-01-30
Fix EC rate in planner vs actual vessel


syscryp:
I'm using the latest dev build. The rate of EC consumption is much higher on active vessels than predicted by the planner in the VAB. 

Unfortunately, the in-flight GUI doesn't provide a breakdown of where the consumption is coming from, so I can't see if it's a particular part. 

Can others replicate?

PiezPiedPy:
That's usually caused by a mod conflict, main suspects are the antennas and pods. If I ever get the time I might update the monitor to show the extra details if its possible.

steamp0rt:
Are you using Module Manager 4.x? If so, downgrade to 3.x.

syscryp:
I was using MM 3.1.3 at the time. I've now upgraded to Kerbalism 2.1.2 and the latest MM. Same problem.

I first encountered the problem with a vessel using a BDB Leo/Gemini pod. I tested it with a stock Mk1 pod as well, same problem. 

The total EC consumption seems to be roughly double what the Planner predicts it should be - but only when there is a Kerbal in the pod. 

I suspect it might be to do with the climatisation process, for two reasons. First, because the problem only arises when there is a kerbal in the pod, and second because climatisation is the only thing I can't turn off using the Kerbalism in-flight UI.

PiezPiedPy:
@HaullyGames or @Gordon-Dry seem to be the most knowledgeable regarding BlueDog support, I'm pretty sure Gordon was also having high EC usage problems but cant remember if he solved the problem.

PiezPiedPy:
@syscryp In case you don't know the support file for Bluedog is in the GameData/Kerbalism/Support/Bluedog.cfg file

syscryp:
Thanks, but like I said, I'm also getting this on stock parts, so I don't believe it's a compatibility issue with BDB.

HaullyGames:
I will review it!
Could you give me a vessel template?
Such as a small mod list + vessel description with small part list.

Note: One problem that I know is related to solar panel, when you load it you have the real amount of EC based on light received for each panel, unloaded vessel most time generates more EC because don't check it, not sure if it is related to my code or it is by KSP design.

Could you confirm if that it is not the case?

syscryp:
[...]

PR #347, opened 2019-03-20
Processes executed 1+n times for one vessel

SirMortimer:
There is an issue with the way Kerbalism generated the vessel IDs it uses for internal reference. 

The problem is with this bit:

`return BitConverter.ToUInt32( v.id.ToByteArray(), 0 );`

v.id (Vessel.id) is a GUID, GUID.ToByteArray() returns 16 bytes. But BitConverter.ToUInt32 uses only 4 of them to get an uint32. This can possibly cause ID collisions where any number of seperate vessels with different GUIDS end up having the same uint32 ID for Kerbalism. Considering that not only vessels are "vessels", but also EVAs, debris and asteroids, we could end up in a situation where an ID-conflicting asteroid and one piece of debris cause the processes for a vessel to be executed, resulting in the process running at 3 times the desired speed (1x for the vessel, 2x for the conflicts).

This might be the cause for #313 but I'm not entirely sure. However I seem to be unable to reproduce that bug any more.

There is a second possibility I can think of that would cause #313, which is that KSP itself has an issue with FlightGlobals.Vessels, resulting in one vessel appearing more than once in that list. I'm not sure if that is the case, indeed I'm sceptical, but none the less I added some test code that will trigger an error message and a log line if it happens. I suggest to keep that in there for a while and see if people report seeing it. If it never pops up, that code can safely be removed again.

### DIFF
--- a/src/Kerbalism/Cache.cs
+++ b/src/Kerbalism/Cache.cs
@@ -9,7 +9,7 @@ namespace KERBALISM
 
 	public sealed class Vessel_info
 	{
-		public Vessel_info(Vessel v, uint vessel_id, UInt64 inc)
+		public Vessel_info(Vessel v, String vessel_id, UInt64 inc)
 		{
 			// NOTE: anything used here can't in turn use cache, unless you know what you are doing
 
@@ -129,7 +129,7 @@ namespace KERBALISM
 		public bool is_vessel;              // true if this is a valid vessel
 		public bool is_rescue;              // true if this is a rescue mission vessel
 		public bool is_valid;               // equivalent to (is_vessel && !is_rescue && !eva_dead)
-		public UInt32 id;                   // generate the id once
+		public String id;                   // generate the id once
 		public int crew_count;              // number of crew on the vessel
 		public int crew_capacity;           // crew capacity of the vessel
 		public double sunlight;             // if the vessel is in direct sunlight
@@ -178,7 +178,7 @@ namespace KERBALISM
 	{
 		public static void Init()
 		{
-			vessels = new Dictionary<uint, Vessel_info>();
+			vessels = new Dictionary<String, Vessel_info>();
 			next_inc = 0;
 		}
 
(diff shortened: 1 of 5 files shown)


lordcirth:
Are GUIDs generated randomly? If they are, a collision in a 32bit space would be unlikely to happen often.

SirMortimer:
I'm just trying to figure that out.

A C# GUID looks like this: `0f8fad5b-d9cb-469f-a165-70867728950e`
Since we only used the first 4 bytes, I'm especially interested in how the first part is being generated.

SirMortimer:
This might be of intereset: https://devblogs.microsoft.com/oldnewthing/20080627-00/?p=21823

lordcirth:
Oh, so depending on the endianness of that timestamp, Kerbalism might have been essentially grabbing the date as it's ID.

SirMortimer:
According to https://tools.ietf.org/html/rfc4122#section-6 the first part is just part of a time stamp.

lordcirth:
My first thought is to just hash the GUID, then truncate the hash. That will give unique IDs.

SirMortimer:
hashing doesn’t guarantee uniqueness.

however we could just use the string representation as ID. i’m a java guy, it’s no issue there, i guess it’s possible in c#

lordcirth:
Given that GUIDs are actually unique, a cryptographic hash would give the full 32 bits of entropy, and I don't see that colliding. If we need better than 32bits, then we can't use uinit32's.

SirMortimer:
We don't have to use uint32, we can use anything. String works just as fine :)

PiezPiedPy:
Just a note on using strings for comparisons, they are expensive, basically the bit size of a CPU is a multiple of 8, chars are 8 bit wide, wchar (Unicode) are 16 bit. Where I'm going with this is that the CPU CMP instruction can only compare as many bits as it is designed for i.e. a 32 bit CPU will compare 32bits in the one instruction cycle or 4 chars or 2 wchars, a 64 bit obviously twice as much.
On a 32bit machine comparing a 128 bit GUID is going to take 4 times longer than an Int32 to complete, which can become an issue when the compare is done multiple times and/or multiple places in a single frame.

A single vessel might not be an issue but with multiple vessels it might.

As a general rule string comparisons should be avoided at all times or kept to a minimum of 4 chars or 2 wchars, unless obviously you have no choice as in searching names for say a contacts list. The same goes for using strings in switch() case: statements, even though such statements are usually converted to Jump Tables by the compiler, using strings in them will destroy any performance gains.

SirMortimer:
OK, then I'd suggest to try with Vessel.persistedId and hope it's unique. If it isn't, we will see error messages pop up in the GUI, so we would notice pretty soon. Let me revert that last commit.

PR #353, opened 2019-03-21
Fixed process running presets in the VAB

SirMortimer:
Some processes (like scrubbers etc.) are meant to be preset to true in the VAB, #338 broke this. I've made a persisted member variable private in #338, looks like that broke this preset.

I guess somewhere this is being set via reflection, I just can't figure out where. Shouldn't have emptied that big glass of scotch yesterday...


lordcirth:
The fact that "Persistant" is mispelled annoys me, but it's probably not worth changing.

SirMortimer:
Same commits are in the "Flags" PR, and this was pushed from my master, so I'll close this one to keep things clean.

PR #357, opened 2019-03-25
I don't see the relation between mass and MTBF

SirMortimer:
And I think it's not the right way to handle this. If the MTBF is wrong in certain parts,
I'd suggest fixing the errors there.


steamp0rt:
lgtm

PR #358, opened 2019-03-26
Support B9PartSwitch (this is purely optional)

SirMortimer:
B9PartSwitch allows to change tank content types directly in the part menu, which is less cumbersome than having to open a new window to do it. This config file replaces Kerbalism's Configure with ModuleB9PartSwitch if B9PartSwitch is installed.


SirMortimer:
aaaah damnit an unclean branch again

PR #365, opened 2019-03-27
Experiments on steroids

SirMortimer:
Added doc for experiment module. Added a bunch of extra experiment restrictions.


steamp0rt:
You don't need to squash commits, I do that on merge

SirMortimer:
I had a bit of a discussion with git, we had a disagreement on which commits actually had to be there.

Git won.

steamp0rt:
Git _can_ stand for "goddamn idiotic truck of shit.

SirMortimer:
Bildschirmfoto 2019-03-28 um 01 14 10
Bildschirmfoto 2019-03-28 um 01 03 47
Demo of limited storage: https://www.youtube.com/watch?v=blAAnM3hIWY

lordcirth:
This looks amazing. What will be the standard way to add more storage? A hard drive part? Config in VAB for extra money? Just higher tech pods?

GameData/Kerbalism/Settings.cfg
```diff
@@ -3,8 +3,6 @@ Kerbalism
   // profile used
   Profile = default                   // valid values are: 'default', 'classic, 'none' or custom profiles
 
-  CrewMass = 0.09375
```
Rockstar04: Is there a reason this can not stay a configurable value? Mods like realism overhaul may want to update this with a number based on human weights.

SirMortimer: I wanted to use a value from PhysicsGlobal, but that seems to be zero.

it won’t stay that way it is, it’s work in progress. but rest assured i’m not going to hardcode any values.

Rockstar04: Absolutely understood that this branch is a WIP, just wanted to chime in before anything got built around this that could make it harder to change later.

Rockstar04:
Can I suggest adding `.DS_Store` to your .gitignore file? There are several that were committed somewhere in this branch.

SirMortimer:
Wait that didn't make it in? I thought I had committed that a while ago

Rockstar04:
I was also surprised with how extensive and well commented the .gitignore file was.

SirMortimer:
Collecting science on the first suborbital flight into space...
Bildschirmfoto 2019-03-31 um 00 40 34
Bildschirmfoto 2019-03-31 um 00 40 45

lordcirth:
So, as experiments are done and data builds up, it will automatically get transmitted, right? So if you had an antenna that could keep up, it would continuously stream the data as it came in?

Issue #356, opened 2019-03-25
Processes run on vessels that shouldn't have them


SirMortimer:
Not sure how to reproduce: some processes will run on vessels that don't have a part for them. For instance you could have a waste processor running without having a chemical plant set to waste processing, a life support module could run a monoprop+o2 fuel cell without having a configuration for it. This is game breaking, it can easily kill a crew.

I *think* the ghost processes that are added all are processes that are unlocked later in the tech tree, but a quick test didn't reproduce the issue. It might be required to restart (!) KSP at times to trigger this bug.

SirMortimer:
I've been looking at this issue and the code that shrouds it for a while, and here are my findings:

1. When you start a new career and launch a single naked Mk1 pod with nothing attached at all, it comes with no less than 7 (!) process controller modules. There is one process controller module for each process that a pod can run at the end of the tech tree. So even if you don't have researched anything yet, you already get the process controller for a O2+monoprop fuel cell that will be researched later. Never will there be a situation in the game where more than at most 3 of those controllers will be in use, but still there are 7 modules that all use memory and processor time.

1a. On a side note: why are there so many air pumps everywhere? It's a space craft, not a bike. Every habitat, every pressure controller, the green house and probably then some modules all contain an air pump which, for 99.9% of all craft, will never be of any use because it only works inside breathable atmosphere which you only get on Duna and Laythe. Why do we have to haul multiple air pumps per vessel everywhere from Moho to Eeloo?

2. Along with the 7 process controllers in that tiny Mk1 pod comes a Configure module that has in its cfg line a mystical incantation which, according to the runes encoded therein, is supposed to set up the process controllers in the way that it was set up during configuration in the Editor. It also adds and removes resources to a vessel, depending on... conditions. This happens at runtime only, when looking at the save file there is no way you can tell which processes are running on a ship unless you look at those runes (an ASCII-length-encoded list of strings). You can't even tell which process controller is supposed to run which process, because they all look exactly the same (!) despite having specific resources associated with them in the profile configuration. Somewhere that specific configuration is removed from the process contorller (!) at runtime. Configure uses profile configuration gathered at KSP startup to do it's magic, and depends on that configuration still matching the vessels (although I'm not certain to what degree this might be a problem). If you want to learn how all of this is happening exactly, you have to follow the module code all the way through. And boy, it's a lot to follow through there. That code is right at the border to incomprehensibility, erring on the wrong side of that border.

[...]

PR #280, opened 2018-11-04
Simulator updates needed for near future electrical

madman2003:
This is all of the work for done to simulator.
This supports per part resources in case they are non-flowable.

PR #371, opened 2019-04-05
Revert part processing introduced with PR #280

### LINKED ISSUE #333: Not sure if this is a bug: "Exception: psuedo-resource _HyLiquefaction must be NO_FLOW"
Kerbalism version: 2.1.2
KSP version: 1.6.1
Module Manager DLL: 4.0.2

Sorry, I don't really use GitHub but thought you might want this in here vs. just having it on the Kerbalism KSP forum thread. Please let me know if you need other information or if this isn't the right place to put this info.

#################
Exception code that is spamming in my latest log:
`[EXC 11:08:08.378] Exception: psuedo-resource _HyLiquefaction must be NO_FLOW﻿
	KERBALISM.Modifiers.Evaluate (.Vessel v, KERBALISM.Vessel_info vi, KERBALISM.Vessel_resources resources, System.Collections.Generic.List 1 modifiers, .Part p, .ProtoPartSnapshot pp)
	KERBALISM.Process.ExecuteVesselWide (.Vessel v, KERBALISM.Vessel_info vi, KERBALISM.Vessel_resources resources, Double elapsed_s)
	KERBALISM.Process.Execute (.Vessel v, KERBALISM.Vessel_info vi, KERBALISM.Vessel_resources resources, Double elapsed_s)
	KERBALISM.Profile.Execute (.Vessel v, KERBALISM.Vessel_info vi, KERBALISM.VesselData vd, KERBALISM.Vessel_resources resources, Double elapsed_s)
	KERBALISM.Kerbalism.FixedUpdate ()
[ERR 11:08:08.394] Resource System Error: Requested resource (id: 1203312488) does not exist﻿`

#################
Here's the full log for reference: [KSP 031219 Time 1111.log](https://www.dropbox.com/s/mcv2fck6zit5j4m/KSP%20031219%20Time%201111.log?dl=0)

SirMortimer:
Reverted the no-flow restrictions introduced by 025c1cb822bf8d58eb99776e123dbb9ec2ca0fee

Process simulation will now be done vessel-wide without a relation to a part. This is how it was before that commit. I somewhat get why the per-part simulation was done, however I don't think that it should ever have been introduced. From what I gathered it was to support one mod, but turned the entire simulation upside down and caused a series of issues in seemingly unrelated places. And why a pseudo-resource needed to be no-flow eludes me.

If that mod (near future electrical) needs special handling, it has to be done differently, maybe by adding new part modules with some sort of special handling.

This fixes #333
This probably breaks parts of Near Future Electrics in some way
This might be related to #356, but not sure


SirMortimer:
As far as I can tell, and please correct me if I'm wrong, this per-part processing was introduced just to simulate a process that cannot interact with resources that are stored in a different part. Uranite cannot flow from a tank at one end vessel to the reactor that sits at the other end of the vessel, and PR #280 was done to simulate that.

And that's the extend to which NFE support would "break". Most people won't even notice, and all but a few won't care even if they do.

PR #423, opened 2019-06-09
Tilted belts

SirMortimer:
Added `geomagnetic_pole_lat` and `geomagnetic_pole_lon` to the radiation model. The magnetic field (and with it the radiation belts) will be oriented according to that pole - which can be everywhere, including the equator...


SirMortimer:
Works for me™

PR #431, opened 2019-06-17
Support for SampleReturnCapsule

Breach-Candy:


SirMortimer:
I think the file name should have the .cfg ending

Breach-Candy:
> I think the file name should have the .cfg ending

I disagree.

SirMortimer:
Where in the tech tree is the sample return capsule unlocked? If it is too soon, the 1MB hdd space might be a bit OP and people might use it instead of, say, the sputnik.

Other than that it looks good to me, waiting for @UsernamesAreAlwaysTaken to make a review (whee, I did't even know that's possible on github :P)

Breach-Candy:
unlocks in survivability. same place as stayputnik iirc

SirMortimer:
if the storage capacity is offset by a higher mass, i'm ok with that.

Breach-Candy:
I'll have to look into it, don't have ksp set up just yet

Breach-Candy:
@SirMortimer alright, had a look into it and it unlocks in landing. Weighs about 100kg. Form factor is the same as stayputnik, only one node on the bottom.

SirMortimer:
One last thing: please add a line to CHANGELOG.md that describes what you've done :)

PR #455, opened 2019-07-05
Apply artificial radiation sources after Gamma Transparency is applied

sswelm:
Currently, any radiation emitters active in the atmosphere have virtually no effect on Radiation while being in a thick atmosphere. For engines like Nuclear Engines specialized for atmospheric usages, like Nuclear Ramjet, Nuclear Turbojet makes no sense the distance is usually very short and due to the effects of neutron scattering, the presence of an atmosphere would rather increase the radiation effect on crew then reduce it.


theJesuit:
Could this be extended to then apply a distance modifier to the centre of the habitable space?

sswelm:
Yes, but this should be part of the Emitter logic. For KSPIE I actually calculate the emitter radiation based on the distance and presence of radiation blocking resources.

SirMortimer:
Thanks!

PR #456, opened 2019-07-07
MoistAtmosphere density fix

Gordon-Dry:
see https://github.com/Standecco/ROKerbalism/pull/10#issuecomment-509024275

> According to the resource file , MoistAtmosphere supposededly has a density of 0.000804. I notice that is very close to the density of Water which is 0.001. Saturated water vapor at STP has a true density of 4.85 g/m3 at 273 K , which translates into a density of 0.00000485 which is 165.77 times less than currently!!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,7 @@
 
 ### Changes since the last release
 
-* Changed density of MoistAtmosphere to 0.00000485 inspired by sswelm aka. FreeThinker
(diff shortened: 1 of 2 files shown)


gotmachine:
I think this is intended, and the rates seems to have been calculated according to this specified density.
See [this comment from the original PR](https://github.com/Kerbalism/Kerbalism/pull/11#discussion_r187796204)
Honestly I recently took the time to look at the whole humidity implementation and I believe the concept is flawed, not to mention the code for it is a ugly hack.
But it seems to be working as intended without major bugs, so I strongly advise not to alter anything.

PR #471, opened 2019-07-19
Smart transmission

SirMortimer:
When transmitting data, prioritize the file with the highest science value per byte


SirMortimer:
"works for me" (tm)

PR #473, opened 2019-07-23
Auto-assign reference bodies for radiation calculations.

Shiolle:
Added automatic reference body assignment for radiation calculations to support systems with multiple stars.


SirMortimer:
Thanks, that actually is a good idea.

Note that we're currently improving multi star support so that twin star systems (like Beyond Home) work better/more accurately. Lib.GetSun(body) will probably be changed to something that returns the celestial body at the center of the planets orbit (which might be a pseudo-body representing the barycenter of a twin star system), so you still might want to configure the reference body by hand to whichever sun is the one with the stronger solar wind. Unless you're OK with the fact that the magnetosphere will orient itself towards the barycenter, which is wrong but not by much.

Issue #439, opened 2019-06-21
StackOverflowException with CommNet Constellation


SirMortimer:
From the forums, https://forum.kerbalspaceprogram.com/index.php?/topic/172400-140-171-kerbalism-v302/page/74/&tab=comments#comment-3621354

> When CommNet Constellation is running with Kerbalism, anytime I view the tracking station or launch another vessel, the game hard crashes and dumps the error to the log.

```
[EXC 19:27:13.192] StackOverflowException: The requested operation caused a stack overflow.
    KERBALISM.Vessel_info..ctor (.Vessel v, Guid vessel_id, UInt64 inc)
    KERBALISM.Cache.VesselInfo (.Vessel v)
    KERBALISM.AntennaInfoCommNet.Init ()
    KERBALISM.AntennaInfoCommNet.AntennaInfo ()
    KERBALISM.ConnectionInfo.GetAntennaInfo (.Vessel v, Boolean powered, Boolean storm)
    KERBALISM.ConnectionInfo..ctor (.Vessel v, Boolean powered, Boolean storm)
    KERBALISM.ConnectionInfo.Update (.Vessel v, Boolean powered, Boolean storm)
    KERBALISM.Vessel_info..ctor (.Vessel v, Guid vessel_id, UInt64 inc)
    KERBALISM.Cache.VesselInfo (.Vessel v)
    KERBALISM.AntennaInfoCommNet.Init ()
    KERBALISM.AntennaInfoCommNet.AntennaInfo ()```

dskzz:
I got this error in a playthrough.  It's strange though.   Only happens on a single design, a probe with those 2 cloverleaf relays, a dish relay, a SLIM probe core and on the lifter an octo core with that sideways communicatron.  I tried launching the same craft in a game with stripped down mods and in sandbox mode in the heavy modded version that crashed in career mode.  It did not crash in sandbox mode.  I'd be happy to send you the save file and the details if you want!

gotmachine:
Following https://github.com/Kerbalism/Kerbalism/commit/1f108c1a0bb305f44291e366a5ce09f52da57946   I think this is fixed in the dev branch as a side effect of the vessel_info refactor, since vessel_info has been merged in VesselData.

The root cause of this is 
- ConnectionInfo.Update request vessel_info
- vessel_info ctor is called and call ConnectionInfo.Update
- wich in turn call the vessel_info ctor
And so on until a stackoverflow exception happen

[...]

PR #489, opened 2019-08-15
Fix for the stack overflow cache recursion loop -  #439

dskzz:
Small change, its a hack fix - does not address the underlying condition causing recursion, just basically saves the last vessel id and if it sees that it's calling function for same id then it doesn't go to cache.  Else proceed normally. 

 #439 was killing me, causing ctd , sometimes without logging the error. Have not had problem since.  Might cause some hinkyness in transmission of science, but not sure of that.

### DIFF
--- a/src/Kerbalism/Comms/AntennaInfoCommNet.cs
+++ b/src/Kerbalism/Comms/AntennaInfoCommNet.cs
@@ -181,6 +181,10 @@ namespace KERBALISM
 			return transmitters;
 		}
 
+
+		private static string lastTestedId;
+
+
 		protected virtual void Init()
 		{
 			if(!antennaInfo.powered || v.connection == null)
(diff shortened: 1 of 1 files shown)


gotmachine:
Actually I think this is fixed in the dev branch as a side effect of the vessel_info refactor, since vessel_info has been merged in VesselData.

The root cause of this is 
- ConnectionInfo.Update request vessel_info
- vessel_info ctor is called and call ConnectionInfo.Update
- wich in turn call the vessel_info ctor
And so on until a stackoverflow exception happen

Since now (in the dev branch) VesselData is never destroyed like vessel_info (it was recreated on-demand), this should not happen anymore.

@dskzz We will soon provide test builds of the dev branch, when we do could you playtest it to check if the exception happen still ?


Please do not not merge this currently.

dskzz:
Yes def let me know I'll dig right in.  Glad to hear this is patched with a refactor instead of my hacky hack.

gotmachine:
@dskzz check https://github.com/Kerbalism/DevBuilds/releases

gotmachine:
Apparently fixed. Just in case, I will let the issue #439 open with the "fiw in dev" tag

PR #505, opened 2019-09-05
improved engine rated ignitions formula

valerian:
Engine rated ignitions are now calculated using a formula that better suits gameplay and common sense:
- the lower the thrust, the higher the ignition count (exponential-ish scale).
- the higher the difference between vacuum and atmosphere ISP, the higher the ignition count.

Meaning that tiny engines always get many ignitions, and the biggest engines get very few, unless they are designed for vacuum use (very low atmosphere ISP compared to vacuum ISP).

See code comments for more details and examples.

### DIFF
--- a/GameData/KerbalismConfig/System/Reliability.cfg
+++ b/GameData/KerbalismConfig/System/Reliability.cfg
@@ -478,19 +478,82 @@
 	}
 }
 
-// this calculates ignitions from thrust. the lower the thrust, the higher the ignition count.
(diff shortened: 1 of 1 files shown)


valerian:
Tested with both DLCs, ReStock, and ReStock+, giving satisfying values.

I removed the x4 multiplier for `@PROPELLANT[XenonGas]` since the new formula gives the ion engine a lot of ignitions (about 200).

lordcirth:
Seems like a pretty solid heuristic to me!

PR #507, opened 2019-09-12
Made public fields volume and surface persistant to allow variable co…

sswelm:
…nfiguration

### DIFF
--- a/src/Kerbalism/Modules/Habitat.cs
+++ b/src/Kerbalism/Modules/Habitat.cs
@@ -7,8 +7,8 @@ namespace KERBALISM
 	public class Habitat: PartModule, ISpecifics
 	{
 		// config
-		[KSPField] public double volume = 0.0;                      // habitable volume in m^3, deduced from bounding box if not specified
-		[KSPField] public double surface = 0.0;                     // external surface in m^2, deduced from bounding box if not specified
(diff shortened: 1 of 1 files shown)


sswelm:
I need it in order to allow switching of habitat configuration, allowing you to sacrifice habitat for more storage of other resources. I also need it for parts that increase in volume after fully constructed

gotmachine:
Well, that won't work. All habitat resources/pseudo-resources amounts/capacities will stay the same, resulting in all derived calculations being wrong.

> I also need it for parts that increase in volume after fully constructed

I think you can use the `state` KSPField for enabling/disabling. Set it to `disabled` in the part config, then call the `Toggle()` method at runtime. Might work, might not. And set `toggle = false` to prevent the player from being able to enable/disable it.

But as @SirMortimer said, I recommend not just not doing any integration with habitat. It has many bugs, won't play nice with any attempt to hook into it from an external plugin, and will probably be rewritten from scratch soon.

sswelm:
Not if you call OnStart after setting the volume, surface and state, then it works as intended. I just want to have the fields persist to make it also function offscreen

gotmachine:
Nope. All "technical" resources capacities won't be updated (Atmosphere, WasteAtmosphere, MoistAtmosphere and Shielding) because Configure() will run only on the first OnStart() call.

SirMortimer:
Habitat is going to be rewritten, probably from scratch. I strongly suggest not to use the current part module.

Issue #520, opened 2019-10-17
Stockalike Station Parts Redux inflatable habitats/centrifuges are not configurable when pod upgrades are enabled by default in sandbox.


DeliciousPi2745:
Occurs on both the latest dev build and version 3.0.2 of Kerbalism, with version 1.2.1 of SSPXr, on Windows 64-bit KSP 1.7.3. The issue can be reliably reproduced with only Kerbalism, SSPXr and their dependencies.

When "pod upgrades installed by default" is enabled in the options when starting a new sandbox save, inflatable habitats & centrifuges introduced by SSPXr cannot have their life support systems configured. They can still be inflated/deployed, have their shielding adjusted, and serve as habitats. All non-inflatable habitats (eg. command modules, hitchhikers) can have their life support slots configured as normal. When the pod upgrades option is left off when creating the new game, the inflatable habitats are configurable as normal.

I've attached my game's output log, please let me know if there's anything else you need :)

[output_log.txt](https://github.com/Kerbalism/Kerbalism/files/3737502/output_log.txt)

PR #522, opened 2019-10-25
Reversed slot upgrades in SSPXr inflatable habitats

lampeh:
The inflatable habs start out with 2 life support slots which get "upgraded" down to 0. This patch reverses the numbers. ( #520 )


gotmachine:
Silly mistake indeed. Thanks for the contribution.

PR #541, opened 2019-11-19
Bluedog_ DB science rework

Gordon-Dry:
I tried to get more out of the patches but made them simpler again before opening this PR.

This means that all probe parts got all the basic experiments - instead of a few chosen ones.

For a better pre-choice for specific probes I suggest that dedicated patches for each probe part are created later.


Gordon-Dry:
Damn it - it just started krakening again.

So I have to use the unfixed parts with obsolete duplicate stock experiment modules.
Okay then.

So if anybody else dares to fix the Bluedog_DB parts regarding Kerbalism science - do it.

PR #549, opened 2019-11-27
Fix slots = 0 in upgrade nodes

Gordon-Dry:


SirMortimer:
Slots = 0 will be overwritten later in the file: 

https://github.com/Kerbalism/Kerbalism/blob/c5bdcb83efb9ade95f0ef4b7596906b7e854be99/GameData/KerbalismConfig/Profiles/Default.cfg#L1718-L1727

And if you look in ModuleManager.ConfigCache, you will see the result, looking like this: ```MODULE
                {
                        name = Configure
                        title = Pod
                        slots = 2
                        UPGRADES
                        {
                                UPGRADE
                                {
                                        name__ = Upgrade-Slots
                                        techRequired__ = electronics
                                        slots = 3
                                }
                        }
```

Gordon-Dry:
That means that the entries with no
slots = X
at all will stay without slots?

Because the patch in Default.cfg does @slots +=
So the entry need to exist before that patch.

Gordon-Dry:
Well, I got it.
But still, the other commits are not affected and you also closed them.
So I need to open a new PR then.

PR #550, opened 2019-11-28
Add support for HabTech2

Gordon-Dry:
Needs testing, but here it is.


SirMortimer:
Never tried it, but I trust this works :)

PR #575, opened 2020-01-06
Duplicate RTG process due to US2 support patch

lampeh:
KerbalismConfig/Support/US2.cfg adds a duplicate _RTG ProcessController module to the **stock** RTG. The part name should be USRTGWedge instead.


SirMortimer:
Thanks!

PR #577, opened 2020-01-09
Fix automation and planner UI for Kerbalism (non-stock) harvesters

asazernik:
In March 2018, 57977b30 (with the uniquely uninformative commit message "Weewoo") deleted Harvester.cs and all references to it.

Assorted commits tried to undo this, notably f17b3892 ("I have no idea how an entire module goes missing") and da1d2dc4 (which fixed background processing). However, the code which supported the Harvester module in the automation system and in the planning UI was never restored. This branch fixes those.

### DIFF
--- a/GameData/Kerbalism/Localization/de.cfg
+++ b/GameData/Kerbalism/Localization/de.cfg
@@ -556,6 +556,7 @@ Localization
     #KERBALISM_Harvester_running = running // UNTRANSLATED
     #KERBALISM_Harvester_stopped = stopped // UNTRANSLATED
     #KERBALISM_Harvester_none = none // UNTRANSLATED
+    #KERBALISM_Harvester_simulatedabundance = Simulate at abundance // UNTRANSLATED
     #KERBALISM_Harvester_land_valid = no ground contact // UNTRANSLATED
     #KERBALISM_Harvester_ocean_valid = not in ocean // UNTRANSLATED
     #KERBALISM_Harvester_atmo_valid = not in atmosphere // UNTRANSLATED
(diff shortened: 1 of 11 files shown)


SirMortimer:
Hi

What a great commit message, I'm still laughing :D

However, please create the PR against the dev branch, which is where the most recent code base lives. We use master only for released versions (and hotfixes for releases, should they be necessary).

asazernik:
Alright, rebased onto dev, and fixed the build; just have to review the planner calculations.

When I build this, I'm noticing that the planner doesn't even include the harvested materials in its planning, but that's true even though they're used by chemical plants. Is there a way to see those numbers in the UI?

asazernik:
Laugh or headdesk, those are the options :-D

SirMortimer:
> Ah, got to be harvesting something defined as a "Supply" in the profile. Will deal with that tomorrow.

Is this finished?

asazernik:
Ran into some weird (game-breaking) issues when asteroid harvesters are on in the background. Next time I get a good chunk of time on a weekend this should be done.

SirMortimer:
OK, then we'll leave this open for now. Just letting you know, we will release a new version that fixes a nasty crash pretty soon, so don't be alarmed when this won't be in it.

asazernik:
Alright, should be good to go!

The AsteroidDrill weirdness is still there, but that's a pre-existing issue, and the code there has an "untested" comment. Next project, since that's what I'm doing a lot of in my current save :-P

gotmachine:
The `#KERBALISM_Harvester_simulatedabundance` string need to be added to the Local class (in LocalizationCache.cs)
Other than that, didn't test but it looks good.

PR #582, opened 2020-01-19
Config partial localization

tinygrox:


gotmachine:
Just in case, please not that the Process{} `name` values cant be localized currently

tinygrox:
Those are in Rule{}, but I will change it if you insist

gotmachine:
I *think* localizing rules name won't cause issues, so no problem for now.
I was talking about process names (further in default.cfg), you didn't do these.

tinygrox:
Well don't worry, I didn't intend to

gotmachine:
Sorry I've made more changes to kerbalism core localization that will conflict with your latest commits.
I've think I've put the changes you made into my latest commit, so when merging if you get conflicts, it should be safe to keep mine (but check anyway). Also, see the LocalHelpers class for a way to auto-generate the localization files.

PR #583, opened 2020-01-20
NearFuture HDD sizes

d4harp:
- Define HDD size for the new NearFutureExploration probes. The values are simple estimates based on existing values on the stock probes. Feel free to adjust if it feels unbalanced.

- Fix a minor typo in the NearFutureSpacecraft HDD patches.


SirMortimer:
Thanks!

PR #584, opened 2020-01-20
XL3 rover wheel planner support

lampeh:
The XL3 rover wheels use ModuleWheelMotorSteering instead of ModuleWheelMotor. This change adds PlannerController modules to all parts matching wildcard ModuleWheelMotor*


SirMortimer:
Thanks!

PR #594, opened 2020-01-29
Update zh-cn.cfg

tinygrox:


SirMortimer:
I just did some proofreading and couldn't find any errors. :)

Thanks!

PR #619, opened 2020-03-24
Create ko-kr.cfg

Vannadin:
Korean translation made by Tomato, me, UNIvErSe, and many Korean users who wanted to remain anonymous.


gotmachine:
Can't merge this into dev. And :
- I sense we will likely need a 3.8 update before 4.0 comes out, so better do those changes in master.
Eventually we can create a "dev-v3" branch, that will allow to keep track of what we will need to "forward-port" in 4.0.
- Localization for dev/4.0 has already changed too much, and will continue to change. I would prefer we don't do any localization work in dev until it is more or less stabilized UI wise.

PR #622, opened 2020-04-22
Prevent patching engine reliability if PayToPlay mod is installed

DarthPointer:
Hi there! I have recently started a new (and the first for me) reliability mod called PayToPlay. The main purpose is to make the player pay for engines maintenance. As for now it is in a deep WIP stage (but you can try it out on engines that are provided with configs or make configs on your own). I wish everything to be ready by the time I finish the initial stage of developing the mod. So I start this PR to prevent engines from being patched by kerbalism and "unpatched", causing extra loading time.

Here is a link to the mod repo: https://github.com/DarthPointer/PayToPlay (no forum thread yet).

### DIFF
--- a/GameData/KerbalismConfig/System/Reliability.cfg
+++ b/GameData/KerbalismConfig/System/Reliability.cfg
@@ -444,7 +444,7 @@
 	}
 }
 
-@PART[*]:HAS[@MODULE[ModuleEngines*],!MODULE[Reliability]:HAS[#type[ModuleEngines]]]:NEEDS[FeatureReliability]:FOR[KerbalismDefault]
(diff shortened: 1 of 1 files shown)


SirMortimer:
Two notes:

1. Please create pull requests for the dev branch, we do all current development there. Master is for the latest released version.

2. We do mod-supporting patches in config files that are dedicated to those mods, that way it is a lot easier to find out which mods have support patches, and what they do exactly - an important consideration for long term maintenance. There is already a support patch file for EngineIgnitor that disables engine failures, I updated that and used it as a template for PaytoPlay: https://github.com/Kerbalism/Kerbalism/blob/dev/GameData/KerbalismConfig/Support/PayToPlay.cfg

Closing this PR in favour of that new config file.

PR #645, opened 2020-06-08
Russian localization, about 40% completed 

Argiziont:
Still working with this loc


Unyo13:
I have more complete version. Should I add the file or can I edit by rows?

Argiziont:
> I have more complete version. Should I add the file or can I edit by rows?

As u want,  u could edit my rows.

Romsik788:
I recommend that you translate from version 4.0. For there are many changes.

Argiziont:
> I recommend that you translate from version 4.0. For there are many changes.

If we could get access to translation file we'll separate work and do translation faster and more reliable.

Unyo13:
Here is my version. An important issue - #KERBALISM_UI_dontcare should be translated shorter to fit the window.
[ru.zip](https://github.com/Kerbalism/Kerbalism/files/4747740/ru.zip)

Unyo13:
Found the dev version, seems to be based on my translation. Please, inform me if any changes needed. There is a plenty of hardcoded strings in 3.9.

Romsik788:
@Unyo13 save your strength and nerves for version 4.0 ;) There I think there will be a lot of work. In version 3.9 nothing else needs to be changed.

PR #646, opened 2020-06-09
Fixing engine runtime spent while railwarping & EI and P2P support

DarthPointer:


DarthPointer:
PR is snowballing xD
Added commits fixing patchorder for EI and P2P support.

GameData/KerbalismConfig/Support/EngineIgnitor.cfg
```diff
@@ -1,5 +1,5 @@
 // disable all engine failures
-@PART[*]:HAS[@MODULE[ModuleEngines*]]:NEEDS[EngineIgnitor]:AFTER[Kerbalism]
```
SirMortimer, line 2: KerbalismDefault will disappear with Kerbalism 4.0 (which the dev branch this patch is against)

SirMortimer:
Used the same logic for an early-return instead of doing the engine time count during time warp. Has the same effect plus saves some CPU cycles. Applied to master + dev.

Thanks!

PR #648, opened 2020-06-26
Solar storms bugfix/additions

ryanc55:
This is a bugfix/addition to the solar storm shielding code.  The existing code is producing some odd/unrealistic results which I believe is down to 3 factors.  

1) The bremsstrahlung formula is inverted.  It goes to infinity at 0 shield mass.  
2) Bremsstrahlung is gamma radiation, which has a rad to rem conversion of 1/10 that of SEP.
https://www.nrc.gov/reading-rm/doc-collections/cfr/part020/part020-1004.html
3) source/receiver distance is being determined by a raycast, which is 0 for adjacent parts, making directly attached shielding useless when combined with issue 2.

Fixes:
1) Use exponential attenuation formula for bremsstrahlung generation
2) Divide bremsstrahlung by 10
3) Use minimum source distance of 1m, making the assumption that the kerbals can get at least that far away from a wall.  This still doesn't account for shield thickness, but plays well enough.

I've tested this in RP-1 with Standecco's proposed RP-1 storm settings and it played fairly realistically.  Mars opposition mission was barely feasible with several tons of extra shielding.  Lunar base missions were easy with dedicated proton shielding, and impossible without them.

For reference, I eyeballed the new linear attenuation coefficient from figure 9 here:  https://www.nrc.gov/reading-rm/doc-collections/cfr/part020/part020-1004.html.   Perhaps this should be affected by the shielding difficulty slider?

### DIFF
--- a/src/Kerbalism/Database/VesselHabitatInfo.cs
+++ b/src/Kerbalism/Database/VesselHabitatInfo.cs
@@ -125,17 +125,18 @@ namespace KERBALISM
 					// to many GeV (the fastest particles can approach the speed of light, as in a
 					// "ground-level event"). This is why they are such a big problem for interplanetary space travel.
 
-					// We just assume a big halfing thickness for that kind of ionized radiation.
-					var halfingThickness = 1.0;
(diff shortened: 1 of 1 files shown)


SirMortimer:
Applied these changes to the dev branch, too. Thanks!

PR #655, opened 2020-08-17
Fix wrong railwarp treatment

DarthPointer:


DarthPointer:
Warning: I think it is possible that engines can start running as soon as we leave railwarp. This can cause us discard wrong burntime from the counter. I guess we should force update of lastRunningCheck if going to fast-return.

PR #661, opened 2020-09-06
Enable water recycler process on US2 Water Purifier wedge.

ccaviness:


GameData/KerbalismConfig/Support/UniversalStorage2.cfg
```diff
@@ -179,3 +179,19 @@
...
+@PART[USWaterPurifier]:NEEDS[ProfileDefault]:FOR[KerbalismDefault]
+{
+  !MODULE[ModuleResourceConverter] {}
+
+  MODULE
+  {
+    name = ProcessController
+    resource = _WaterRecycler
+    title = Water recycler
+    capacity = 18 // part is 0.24t, ECLSS is 0.04 for capacity = 3
```
ccaviness, line 192: While this is plenty for all but the biggest uses, I could even see 54 as valid since the water recycler in the ECLSS is one of 3 slots, so ostensibly 0.24t/.(04t*3 slots) * 3 capacity = 54.

PR #664, opened 2020-09-17
crewReport only for crewable parts

Gordon-Dry:
For some reason the 'dev' branch on Kerbalism repos got no System folder, so I had to create one in my branch and add this file ...
Weird.


gotmachine:
Dev is kerbalism 4.0 wich doesn't have a System folder and has almost every config file moved around.

PR #665, opened 2020-09-23
Fix stock bug with module generator in launch clamp and is always active = false

Gordon-Dry:
The is no button to toggle on the PAW so when it's off it's off and so the planner button to toggle consideration is useless.

Solution:
always active = true

Modules:
LaunchClamp
ModuleRestockLaunchClamp

which got a ModuleGenerator.

### DIFF
--- a/GameData/KerbalismConfig/Support/ReStock.cfg
+++ b/GameData/KerbalismConfig/Support/ReStock.cfg
@@ -15,4 +15,10 @@
 	{
 		%isAlwaysActive = true
 	}
+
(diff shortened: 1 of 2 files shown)


Gordon-Dry:
Hold on, while doing A I stumble upon old unidentified issue B
very kerbal...

Gordon-Dry:
It's so obvious - I'm fiddling and trying and ... but the culprit is a bug in the code:
https://github.com/Kerbalism/Kerbalism/issues/666

Gordon-Dry:
Closing PR in favor of https://github.com/Kerbalism/Kerbalism/issues/666

PR #675, opened 2020-10-03
fix SSTU.cfg

Gordon-Dry:


gotmachine:
Dev is Kerbalism 4.0 which has a completely different config format.

PR #677, opened 2020-10-04
add support for Luciole

Gordon-Dry:
Need some help about how to convert the config syntax from Kerbalism 3.x to 4.x


gotmachine:
Dev is Kerbalism 4.0 which has a completely different config format.

PR #691, opened 2020-11-14
French Localization Support

OnlyLightMatters:
2 new files for French.
Some texts are still not translated in the UI but this already better than nothing.
Working on understanding what more should be done.


GameData/Kerbalism/Localization/fr-fr.cfg
```diff
@@ -0,0 +1,1146 @@
+//////////////////////////////////////////////////////////////////////////////////////
+// NEVER EDIT THIS CFG FILE MANUALLY, IT IS AUTOGENERATED.
+// CHANGES MUST BE DONE IN THE CODE, INSIDE THE `Local` CLASS (LocalizationCache.cs)
+//////////////////////////////////////////////////////////////////////////////////////
+
+Localization
+{
+  en-us
```
gotmachine, line 8: Change to fr-fr

GameData/KerbalismConfig/Localization/fr-fr.cfg
```diff
@@ -0,0 +1,560 @@
+Localization
+{
+	en-us
```
gotmachine, line 3: Change to fr-fr

gotmachine (commented):
Thanks for the submission.

This said, the localization is currently quite a bit of out of sync with the codebase, as we are changing a lot of things internally, and there will possibly be a lot more changes if we do the UI rewrite.
Since syncing the localization files is a bit tedious, we are waiting to have a somewhat stabilized codebase before properly cleaning it.
Also, there isn't a lot of involvement currently, and frankly we have no idea when the 4.0 (dev branch) will turn into a release.

So, not to discourage you, but just be aware that there are a few caveats to doing localization work in the current state of affairs.

GameData/Kerbalism/Localization/fr-fr.cfg
OnlyLightMatters: Thx for the update.
I don't mind working on a new translation later.

GameData/KerbalismConfig/Localization/fr-fr.cfg
OnlyLightMatters: Thx for the update.

PR #724, opened 2021-03-22
Relicense to 0BSD

aspenluxxxy:
[The Unlicense is not entirely valid or legal everywhere, mainly Germany](https://softwareengineering.stackexchange.com/questions/147111/what-is-wrong-with-the-unlicense)

[0BSD is a license that effectively allows the same things as the Unlicense, while being actually legally valid](https://opensource.stackexchange.com/questions/7713/is-there-any-reason-to-not-use-0bsd).

(I _am_ N70, by the way. Check my forum profile, it links to my twitter, which links to this github account)


gotmachine:
Spoiler alert : every "we relinquish all rights" license that exist isn't legally valid and doesn't protect anybody from someone with the will and enough resources to contest it in a legal context.

And I like the idea of of the unlicense, which has a slight "put your copyright system up your a**" tone to it. This is a badly written mod for a niche game, relax and enjoy.

Frankly, I absolutely don't care, and I won't go through the hassle of updating every single place where we mention the license (forums, spacedock, CKAN...).
And moreover, if I go trough it, I would use the MIT license which is the most widely adopted license for that use case.

PR #756, opened 2021-09-16
Stock, Restock+, RLA probe core data

hypodronic:
Probe numbers based on subtracting the mass attributable to gyros and batteries, then multiplying the remaining mass by the cost and a constant related to the tech tier. Big difference is providing non-basic hard drives for some very high-end probe cores.


gotmachine:
Reducing stock probes data capacity will likely cause issue with existing saves. I advise not doing it, especially since the proposed changes are quite minor.

PR #765, opened 2021-10-29
RemoteTech Kerbalism4 backport

Ko0z:
Backport of the new Comms/Signal handling from Kerbalism4.

*RemoteTech should now be behaving as you'd expect. When the vessel run out of power or when there's radio blackout because of LOS, the vessel can't be modified through the UI

*CommNet has not be thoroughly tested, neither has RealAntennas, will require some testing before release


Ko0z:
I've tested as much as I can and it seems to be working as intended. Let me know if there is anything that I need to do or change etc...

gotmachine:
I will merge this into dev and make a public dev build, so we can get some feedback before pushing this on the usual distribution channels. We at least need confirmation from the RO folk that the RA integration is working as expected.

Issue #779, opened 2021-11-28
Planner works with ModuleGenerator but not with ProcessController (_ECGen)


Gordon-Dry:
A launchClamp that uses stock ModuleGenerator is taken into account by PlannerController (generator) but a launchClamp with ProcessController (_ECGen) is not.

Gordon-Dry:
see
https://github.com/Kerbalism/Kerbalism/pull/780
and
https://github.com/Kerbalism/Kerbalism/pull/781

Gordon-Dry:
Well, for now https://github.com/Kerbalism/Kerbalism/pull/782

gotmachine:
I guess this isn't fixed.

Aebestach:
Pad power from `ProcessController` (`_ECGen`) **is** already included in the Planner (~1 EC/s generator). What’s missing is only a `PlannerController` toggle to include/exclude it.

Wiring that properly needs Planner code changes (processes use vessel-wide pseudo-resources and ignore `PlannerController`), not just cfg. Too much complexity for little benefit.

Workaround: don’t attach launch clamps while checking mission EC in the Planner.

PR #781, opened 2021-11-28
ModuleRestockLaunchClamp as well

Gordon-Dry:
see https://github.com/Kerbalism/Kerbalism/issues/779


Gordon-Dry:
Both methods are borked. Neither the old method nor the new method work, so f it.

Gordon-Dry:
I wanted to say:
In VAB, the PlannerController neither is capable of recognizing a stock ModuleGenerator's EC output nor a Kerbalism ProcessController's _ECGen output - which is weird. At least the latter should work. It's the own league.

PR #785, opened 2021-11-29
Fix antenna EC calculation typo

shult12:
A typo meant that some antennas, for example feedable antennas from Near Future Exploration had too much EC consumption.


Gordon-Dry:
Good catch.

PR #786, opened 2021-11-29
Avoid having unlimited Data storage in HardDrives

Gordon-Dry:
Avoid having unlimited Data storage in HardDrives which should only hold physical samples, like for example the bluedog_GATV_MMDetector


gotmachine:
Seems to be superseded by #797

PR #797, opened 2021-12-25
Patches-HardDrives

Gordon-Dry:
This is already the config from dev, but with a fix in line 488.


gotmachine:
So I guess this replace the #786 PR (Avoid having unlimited Data storage in HardDrives which should only hold physical samples) ?

I'm reluctant to merge this, looks like this could cause side issues and I don't have the will to verify.
Leaving open until someone review and test it and confirm this is doesn't wreak havoc on other drive configs.

gotmachine:
Closing due to lack of activity

Issue #762, opened 2021-10-15
Borked Lab patches


UsernamesAreAlwaysTaken:
Ok.

You may or may not remember me at this point, but I wrote a few lines of patches primarily for science.

Getting back and digging in - I just noticed that the Laboratory patching is borked (I assume someone modified it after my departure).

What do I mean by borked? You have several parts in that patching system that does specific things each pass, and if one step is messed up, a lot of unfunny things happen. Courtesy of myself for writing such an "easy to understand" and "fun" system.

Now, to the problem itself:

Your have this:
@PART[*]:HAS[@MODULE[ModuleScienceLab]]:NEEDS[FeatureScience]
{
	!MODULE[ModuleScienceLab] {}
	!MODULE[ModuleScienceConverter] {}

	MODULE
	{
		name = Laboratory
		researcher = Scientist
		analysis_rate = 0.005 // 5 kbps
		ec_rate = 1.0
	}
}

Which nukes all stock lab modules on everything it finds and replaces it with our own module. Does nothing else.

Then you have this:

@PART[Large_Crewed_Lab]:HAS[@MODULE[ModuleScienceLab]]:NEEDS[FeatureScience]:FIRST
{
	MODULE
	{
		name = Configure
		title = Laboratory Experiments
	}
}

Which adds the functionality to the lab - all the epxeriments, the ability to swap them, etc.

That's awesome, but it's only awesome for the stock Lab. All other labs in the game are stuck with the kerbalism module and nothing else. The Configure (Lab Experiments) module and the experiments themselves never get added to anything else besides the stock Lab.

Now, I'm confused. I assume this was done to reduce the amount of complaints regarding broken mods, such as WBI for example. 
But I don't think this was the way to do it.

Now I can probably do something about it, but at this point I sort of have to ask HOW do you want the fix? Do I leave all other modded labs untouched by Kerbalism, or do I hand-pick which labs get the patches? Third option was to revert to old state, where everything gets patched, but I assume the complaints got too much.

kleinschmidt:
I just hit this with SSPX labs too.  Any reason not to just revert https://github.com/Kerbalism/Kerbalism/commit/5a8840238a7ec08ac65659de97e05f749bffb3a7 ?  I searched for `ModuleScienceLab` in my (lightly modded) GameData and the only places it pops up are the stock lab, the SSPX labs, and some contracts.

gotmachine:
Fixed in https://github.com/Kerbalism/Kerbalism/pull/830

PR #830, opened 2022-09-24
Small mod support changes and fixes

DangerNoodle9000:
(Basicially a copy of my forum post)

I would like to share some changes/additions I've made to a few .cfg files. These include:

**SSPX.cfg:**
- Changed cupola-greenhouse-125-1 from 1/90 to 1/6 of kerbalism greenhouse
- Added custom greenhouse module for aquaculture-375-1

**SSPX_Science.cfg:**
- Added a patch to change the analysis speed of some labs

**UniversalStorage2.cfg:**
- Added support for the US2 sabatier reactor part, scaled accordingly

**Patches-Experiments.cfg:**
- Changed a line to enable kerbalism to add lab experiments to all labs and not just the stock lab

**Patches-HardDrives.cfg:**
- Changed a line to enable the patch to correctly add hard drives to labs

 

**Detailed explanation:**

**SSPX.cfg:**
The cupola greenhouse has 1/12 as many individual plants as the greenhouse-375-1, wich is 2x as good as the default kerbalism greenhouse.

The fish tank is different from all the others. I decided it gets 4x kerbalism greenhouse ("kg") / 2x 375-1 greenhouse food production.
It is now some sort of direct Waste-converter, as i imagine some fish species can use the biomatter contained in waste more efficiently than just extracting the wastes nh3
Uses 2 Kerbals worth of Waste as fish-fodder, plus algae from integrated tanks that are as effective as 1x "kg" in both input and output
The fish however only consume/produce 1.75 Kerbals worth of co2/o2 while ingesting 2 Kerbals of food per day (not actual biomass of a Kerbal, the amount of food eaten by one per day)
I suspect growing lifeforms exhale a lower percentage of their ingested carbon as co2, than fully grown individuals of their kind,
because at least some of that carbon  needs to be integrated into their bodies for growing.

**SSPX_Science.cfg:**
The large labs can fit more scientists, and house more and also probably more modern equipment, so they should be realistically abe to analyse samples faster.

**UniversalStorage2.cfg:**
There was no config for the sabatier reactor so I added one that is scaled in the same vein as the other parts, using its mass in relation to the kerbalism ECLSS

**Patches-Experiments.cfg:**
The line in question effectively adds lab experiments to parts with a kerbalism laboratory module, but only targets the stock lab for whatever reason, despite other lab related patches targeting all parts with lab modules
Now all modded labs too get lab experiments

**Patches-HardDrives.cfg:**
The line in question adds a hard drive modules to all parts with a stock lab module, but that stock lab module got removed two files above.
I've changed the filter module to the kerbalism lab module instead.

### DIFF
--- a/GameData/KerbalismConfig/Support/SSPX.cfg
+++ b/GameData/KerbalismConfig/Support/SSPX.cfg
@@ -1104,7 +1104,6 @@
 	}
 }
 
-// DangerNoodle: 
 @PART[sspx-cupola-greenhouse-125-1]:NEEDS[StationPartsExpansionRedux,ProfileDefault]:AFTER[KerbalismDefault]
 {
 	MODULE
--- a/GameData/KerbalismConfig/Support/SSPX.cfg
+++ b/GameData/KerbalismConfig/Support/SSPX.cfg
@@ -1104,6 +1104,7 @@
 	}
 }
 
+// DangerNoodle: 
 @PART[sspx-cupola-greenhouse-125-1]:NEEDS[StationPartsExpansionRedux,ProfileDefault]:AFTER[KerbalismDefault]
 {
 	MODULE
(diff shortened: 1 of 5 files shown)


gotmachine:
Looks good to me.
For reference, this does what was suggested in https://github.com/Kerbalism/Kerbalism/issues/762

DangerNoodle9000:
awesome!

PR #831, opened 2022-09-28
Localization improvements & update zh-cn.cfg

tinygrox:
😁
Does it look good?


src/Kerbalism/Database/VesselSituations.cs
```diff
@@ -15,11 +15,11 @@ public class VesselSituations
...
 		public List<VirtualBiome> virtualBiomes { get; private set; } = new List<VirtualBiome>();
 
-		public string BodyTitle => body.name;
+		public string BodyTitle => body.displayName.LocalizeRemoveGender();
 		public string BiomeTitle => biome != null ? biome.displayname : string.Empty;
 
-		public string BodyName => body.name;
-		public string BiomeName => biome != null ? biome.name.Replace(" ", string.Empty) : string.Empty;
+		public string BodyName => body.displayName.LocalizeRemoveGender();
+		public string BiomeName => biome != null ? biome.displayname.Replace(" ", string.Empty) : string.Empty;
```
gotmachine, line 22: *Name properties are meant to return the internal (non-UI / localized) name. Will revert that on my end.

tinygrox: sry 🥲

PR #836, opened 2022-11-17
Add celestial body shielding to GammaTransparency

ryanc55:
Partially block GCR based on visual angle of local body.

### DIFF
--- a/src/Kerbalism/Sim.cs
+++ b/src/Kerbalism/Sim.cs
@@ -742,11 +742,12 @@ namespace KERBALISM
 		}
 
 
-		// return proportion of ionizing radiation not blocked by atmosphere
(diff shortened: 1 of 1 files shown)


gotmachine:
While taking body occlusion into account for background radiation is indeed missing, this isn't the right place to implement it.

The GammaTransparency factor is applied to all radiation sources (including belts/magnetospheres), not only background radiation (GCR), so introducing that occlusion factor here doesn't work and will produce unwanted side effects.

It should likely be a separate factor applied only to `ExternRadiation` here : https://github.com/Kerbalism/Kerbalism/blob/f4fccfd195dddfed3a56dbbb9d06892ab86c0fcf/src/Kerbalism/Radiation/Radiation.cs#L746

Closing this, feel free to make a new PR.

PR #843, opened 2022-12-19
GravityRing should use deltaTime, not fixedDeltaTime

JonnyOThan:
This function is called from Update and not FixedUpdate.  Using fixedDeltaTime results in the centrifuge spinning much faster than it should.


JonnyOThan:
Hmm...not sure this is actually correct either.  It seemed to change speeds when I toggled the IVA overlay.

gotmachine:
I don't really have the motivation to test this, in any case this is entirely cosmetic so I guess I can merge anyway ?

PR #846, opened 2023-01-26
Hacky means to fix dumping, review before merge!

R-T-B:
This is really hacky and may even leave some garbage leftover but it does reliably fix the dumping settings issues.  I am reviewing it in my save now to see how well it works over time.

Don't merge just yet, but feel free to compile and try.


R-T-B:
As we discussed, there is a few reasons this does not work.  Closing.

PR #847, opened 2023-01-29
Workaround broken "dump=true" setting

R-T-B:
Dump=true was broken in the recent dump fixes.  Rather than try to figure out why, you can address it config side, as shown here.  Up to you which fix you use.


R-T-B:
got has opted to work around this in code.

PR #877, opened 2023-11-28
Support priority in analytic resource processing (defaults to off)

NathanKell:
Also use resource ID when possible. Note: if we were willing to accept the GC impact of creating wrappers for PartResource vs ProtoPartResourceSnapshot or if we static-allocated a bunch of holders (or if C# supported type-specific generic methods, grr) we could eliminate a bunch of duplicate code between loaded and unloaded vessels.

Given the switch to straight int keys rather than having to gethashcode on strings all the time, my hope is this will be about as fast as before.

### DIFF
--- a/GameData/KerbalismConfig/Settings.cfg
+++ b/GameData/KerbalismConfig/Settings.cfg
@@ -64,6 +64,7 @@ Kerbalism
   // ModsWarning = none
 
   UseSamplingSunFactor = false        // use experimental sunlight factor calculation for fast time warps
+  UseResourcePriority = false         // whether to respect part flow priority when applying resource deltas
   
   // debug / logging
   VolumeAndSurfaceLogging = false      // set to true to have the full output of crewed parts volume/surface calculations written to the KSP.log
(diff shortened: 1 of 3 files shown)


gotmachine:
Nice I guess... Altough I fail to see the point, since we don't track where the producers/consumer actually are...
I guess this somehow make sense in some specific scenario ?

NathanKell:
Yeah this couldn't support crossfeed checks, which is why it treats everything that has priority as STAGE_PRIORITY_FLOW_BALANCE rather than ALL_VESSEL_BALANCE (Kerbalism as it stands treats everything as the latter only).

The relevant scenarios:
* You have a vessel, and you want it to drain EC from lower stages first. This is very important in RO where often you can't generate power yourself.
* You have a 2-stage lander, and you have life support supplies split between the descent stage and ascent stage (consider the historic LM, with most supplies in the descent stage). Or you have a capsule/CSM setup, where resources are split with only a tiny bit in the capsule itself. You never want to have to remember to transfer resources before decoupling.

While we don't really handle ISRU, I could certainly see a case for stock players who'd want background-generated resources to be first put in some tanks and only later other ones.

NathanKell:
Rewrote this to use a static-cached set of wrapper objects that get linked each pass. Also switched the whole ResourceInfo system over to being ID keyed for more speed. This no longer allocs at all when syncing resources (unless the caches aren't large enough yet, and need to be grown).

PR #884, opened 2024-01-03
Fix for sickbays erroneously consuming resources after rendezvous

DougWagner:
This is a fix for a bug I found with sickbays where rendezvousing with a vessel that contains a part with a configurable sickbay would cause it to start consuming electricity and other resources as if all possible sickbay modules were running at the same time. To reproduce the bug, launch a vessel with a PPD-10 Hitchhiker crew container with nothing configured in the sickbay, and then launch a 2nd vessel and rendezvous with the first. When the vessel with the Hitchhiker container comes within physics range and loads into the scene, it will immediately start consuming electricity and oxygen for both the sickbay TV and RDU processes, despite originally being configured with neither sickbay modules. This bug is present in both Kerbalism versions 3.18 and 3.19.

Here are screenshots showing electricity consumption before and after rendezvousing for the target vessel that has a PPD-10 Hitchhiker part on it:

Before rendezvous:
before-rendezvous
After rendezvous:
after-rendezvous

As shown in the screenshots, the vessel has started consuming electricity and oxygen for the RDU and TV, despite being configured with neither when launched.

When debugging this, I discovered that loading a vessel normally by launching or focusing it from the tracking station would call `Sickbay.Start()` immediately followed by `Sickbay.Update()`, however when rendezvousing with the vessel, only `Sickbay.Start()` would be called. If both Start and Update are called, the resulting `amount` values for `_SickbayTV` and `_SickbayRDU` are set to 0, but if only Start is called, the `amount` values would end up being set to 1. This is because `Sickbay.Start()` calls `Sickbay.Configure()` with the first parameter hardcoded to `true`, resulting in `Lib.SetResource()` to be called for the TV and RDU with an `amount` value of 1. I am unsure if rendezvouses only calling `Sickbay.Start()` without calling `Sickbay.Update()` is a separate bug or not.

This change now ensures that the `amount` value for `_SickbayTV` and `_SickbayRDU` will only be set to 1 if the sickbay module configured in the part, which will set `isEnabled` to `true`. Otherwise, if the sickbay is not configured, `isEnabled` for `_SickbayTV` and `_SickbayRDU` will be `false`, and the `amount` value will always be 0.

I have tested this change with various different sickbay configurations and have not found any erroneous behavior.

### DIFF
--- a/src/Kerbalism/Modules/Sickbay.cs
+++ b/src/Kerbalism/Modules/Sickbay.cs
@@ -77,7 +77,7 @@ namespace KERBALISM
 			}
 
 			// configure on start
-			Configure(true, slots, cureEverybody);
+			Configure(isEnabled & running, slots, cureEverybody);
 
 			UpdateActions();
 		}
@@ -122,7 +122,7 @@ namespace KERBALISM
 			}
 			RemovePatients(removeList);
 
-			Configure(running, slots, cureEverybody);
+			Configure(isEnabled & running, slots, cureEverybody);
 
 			if (part.IsPAWVisible())
 				UpdateActions();


willfulZephyr:
This happens for me when EVAing as well. Any ETA on this PR getting merged and released?

willfulZephyr:
> This happens for me when EVAing as well. Any ETA on this PR getting merged and released?

Nevermind - I can't reproduce the EVA thing.

Zondulus:
Hello,

I had this issue occur across 50+ crafts in a 150 hr. career save when updating from 3.17 to 3.18. Like Doug Wagner I also was using SSPX but it happened with the vanilla Hitchhiker as well.

I support Doug Wagner's suggested fixes to this issue! It is a frustrating problem to have a permanent 3.5/s EC drain on your craft. In addition, rolling back to 3.17 does not fix the problem once it occurs.

Westward96:
Greetings,

Can also confirm this is an ongoing bug impacting any module that has the sickbay slot available. Please consider implementing Doug Wagner's fixes.

Monkey1049:
I am also having this problem would be great if it was implemented soon.

PR #914, opened 2025-08-09
Error handling fallback for NaN dampingExponent when using RemoteTech

KerballOne:
There probably is a more comprehensive fix to ensure that dampingExponent is never NaN.  Maybe redoing the SignalStrength so it never takes the LOG(0).  But this simple error check is not likely to break anything else, as it only executes on NaN.

### DIFF
--- a/src/Kerbalism/Sim.cs
+++ b/src/Kerbalism/Sim.cs
@@ -1265,6 +1265,12 @@ namespace KERBALISM
 				dampingExponent = Math.Log(desiredRateAt2AU / baseRate, strengthAt2AU);
 
 				// 2.4 seems good for RemoteTech
+				if (dampingExponent.ToString("F4") == "NaN")
+				{
+					Lib.Log("dampingExponent is " + dampingExponent + ",... setting to 2.4");
+					dampingExponent = 2.4;
+				}
+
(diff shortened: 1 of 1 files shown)


KerballOne:
https://discord.com/channels/418222516386004992/418222516386004994/1402431364715905034

https://discord.com/channels/418222516386004992/418222516386004994/1403956380096008222

R-T-B:
Thanks for this.  I'm running the change through my test install now, and it should make it in next release if it doesn't do any odd behavior.

src/Kerbalism/Sim.cs
```diff
@@ -1265,6 +1265,12 @@ public static double DataRateDampingExponentRT
 				dampingExponent = Math.Log(desiredRateAt2AU / baseRate, strengthAt2AU);
 
 				// 2.4 seems good for RemoteTech
+				if (dampingExponent.ToString("F4") == "NaN")
```
gotmachine, line 1268: Nope.
`if (double.IsNaN(dampingExponent))`

R-T-B (commented):
Looks ok now. I will merge, and test in my live save for about a week as usual. Also working on some config updates for NFE, which will combine with the next release.

Issue #821, opened 2022-07-24
Pressure system is glitchy


gotmachine:
The pressure system is overall very glitchy, leading to several end-user issues. 

This is a meta-issue tracking individual reports :
- #766
- #640
- #598
- #790

The pressure system code and implementation is a total mess and I believe has fundamental issues that can't really be fixed without a deeper rewrite. This will not be worked on by me, so contributions are welcome.

SerShrubert:
Some minor issues I have noticed:

- Vent button is always visible, even in breathable atmosphere. There is no way to disable this in the module like with the toggle button
- Pressure control is disabled in breathable atmosphere, but the Air Pump process only exists in the life support part.
- Because of the above two issues, its possible to vent the atmosphere on the ground to create a vacuum in the habitat, with no way to replenish it.

Can Vent atmosphere be hidden by default? Not sure the use case for this. It clutters the PAW and its easy to accidentally click.
Air Pump could be added to modded inflatable ground based habitats by default without taking up a configure slot.

Also while working on my BDB support patch I was able to get Wet Workshops to work by using the existing in flight part switch to force the habitat state into the "Deploying" state. It feels a little hacky but it works. But since toggle isn't a persistent field I can't change it from a part switch in flight and have to keep it disabled. Could toggle be changed to persistent so this can be changed without altering state directly? Or could a forcedDisabled state be added to hide the toggle until its changed to the regular disabled state. Actually implementing the wet workshop in Kerbalism feels a little out of scope, adapting the existing part switch feels like a convenient workaround.

Issue #766, opened 2021-10-30
[1.12.2]pressurization (pressure control) is not working properly under high warp speeds


tonimark:
when travelling on a LOADED vessel at high warp speeds the habitats are not pressurized properly due to calculation errors on high warping speeds (50x and above) as a result these 2 things are happening:
1. when the pressure reaches nearly 100kpa it slows down and consumes excessive nitrogen(higher warp)
2. pressure level loops between 90kpa and 70kpa(lower warp)
3. if a high pressure of about 100.9 is reached it will remove all pressure going back to 0kpa (when fluctuating between warps)
 AND also even when inflating the habitat when the vessel is unloaded loading it backdrops the pressure down to 70kpa 
all these problems cause constant unrealistic pressure loss and unnecessary nitrogen consumption when inflating habitants 
the problem doesn't occur on lower warp speeds-slower pressurization
replication: build a large vessel (habitat disabled) and launch it into space with a few pressure control modules and then high warp and watch the habitat pressure gauge from the menu under high warp speed

gotmachine:
The entire pressurization system is more or less completely broken (see also #598 and #640), regardless of loaded state or timewarp speed.

I personally have no plans to fix it.

PR #921, opened 2025-08-17
Habitat refactor

gotmachine:
### Overview
There are several bugs related to the Habitat module and its associated functionalities : pressure, living space, radiation shielding. Overall, the code is a huge mess resulting from years of building up upon an inadequate base and slapping poor quality (and/or overengineered) hacks/features/bugfixes on top of that. This PR aims at fixing the various bugs and cleaning/refactoring the whole thing while keeping backward compatibility with existing saves and configs. Functionality and feature set is broadly similar to previous releases, don't expect anything fancy aside from more reliable and logical behavior, and a few small QoL things.

### Changes
- General cleanup of the Habitat module, removed a ton of dead code, various optimizations and bug fixes.
- Removal of the `ResourceBalance` class in favor of a saner, simpler implementation of atmosphere/wasteAtmosphere equalization reusing the existing `VesselHabitatInfo` infrastructure. Equalization is now instantaneous and only applied to fully enabled habitats, it isn't involved anymore in pressurizing inflatables.
- Revamped state machine in the Habitat module, allowing to dissociate inflatables deployed by pressurization from mechanically deployed habs. This is is configured through a few additional config fields, and implemented by the non-inflatable SSPX large gravity rings.
- Restored the ability to disable habitats. Disabling habitats don't trigger a venting of their atmosphere anymore. Disabling is correlated to retracting only if explicity defined in the config.
- Removed/depreciated the not implemented partial pressure thing in favor of a simpler pressurizable/not pressurizable config toggle.
- Implemented a loaded/unloaded agnostic update loop for habitats, reusing as much as possible the existing infrastructure (`VesselHabitatInfo`) / computations (`Background.Background_PMs()`. This now : 
  - Centralize hab pressure equalization
  - Is responsible for providing the volume, surface, pressure, poisoning, etc values used elsewhere.
  - This allow to decouple those stats from the vessel wide resources states, which was necessary to properly handle non-pressurizable habs, as well as using a correct pressure level in inflating/deploying/pressurizing states. 
  - Apply the atmo leaks rule to disabled habitats, enabling the ability to disable them without having to depressurize them.
- Implemented the ability to vent habitats, independently of the enabled/disabled state
- Inflating now requires to reach a pressure level equal to `Settings.PressureThreshold`, instead of full pressure
- Everything Habitat related should behave identically at low and high timewarp rates, as well as on loaded and unloaded vessels.
- Implemented an optional pressure equalization mechanism allowing to speed up pressurization of inflatables / deployables by transferring atmosphere from enabled habitats, and consequently with the caveat that you can end up loosing the pressurized state. The option is only available in the `inflating` and `waitingForPressure` states, and will only show up if it is relatively safe to use (enough pressure in enabled habs, enough nitrogen to repressurize, enough pressurization capacity). Equalization speed can be changed through the `EqualizationRateFactor` setting.

This should fix the various pressure related bugs, see issues #821 and #766

### Config changes :

The changes should be fully backward compatible with existing configs (and saves).

**New fields in the Habitat module**
- `nonPressurizable` (`bool`, default `false`) : if true, the part can't be pressurized
- `inflateRequiresPressure` (`bool`, default `true`) : if false, inflating/deploying doesn't require pressurizing and can be done freely
- `canRetract` (`bool`, default `false`) : if false, the part can't be retracted once inflated/deployed

For deployable habs, the module *should* support any combination of the existing `inflatableUsingRigidWalls` and of the new `inflateRequiresPressure` and `canRetract` fields. Behavior will stay identical as in previous versions, meaning all deployables are deployed by pressurizing them, and can't be retracted afterwards, at the exception of the SSPX 3 large "rigid" gravity rings, which don't require pressure to be deployed and can be retracted (see updated support config).

**Depreciated fields in the Habitat module**
- `max_pressure` (`double`, default `1.0`) : this is kept around for backward compatibility. Having it set to a value below `Settings.PressureThreshold` will cause the `nonPressurizable` field to be set to `true`.

**Atmo leaks process**
The atmo leaks process is now applied "manually" to non-enabled habs still holding atmosphere, from the new habitat-speciific update loop in `VesselHabitatInfo`.
This is required because the process will only consider parts where the `Atmosphere` resource has `flowstate = true`, but we can't do that for non-enabled habs, as they would then be considered by the other processes (notably pressure controllers). There are various options for doing this, and none of them are pretty. In the end, I decided to semi-hardcode the behavior, grabbing the leak rate from the matching profile process. Other "cleaner" options would have required a much deeper refactor and would have broken backward compatibility with existing configs, all for a very minor functionality/feature.

The atmo leaks process to use is identified by 3 conditions :
- Either its name is `atmo leaks`, or the process has an additional `isAtmoLeaks = true` config value.
- It has the `surface` modifier
- It has an `Atmosphere` input

The losses are only applied when the environment isn't breathable (replicating the `non_breathable` modifier), with the following formula :
`double atmoLossesForPart = atmoLeaksProcessAtmosphereInputRate * partSurface`

**Equalization speed**
A new `EqualizationRateFactor` field is available in `settings.cfg`. It controls the equalization rate when manually equalizing an inflatable/deployable hab. Value is in % per second (at max pressure difference). Default value is 0.01 (1%).

### Dev releases : 
[Kerbalism-Core_dev-9363_for_KSP1.8.0-1.12.9.zip](https://github.com/user-attachments/files/21867532/Kerbalism-Core_dev-9363_for_KSP1.8.0-1.12.9.zip)
[Kerbalism-Config_dev-9363_for_KSP1.8.0-1.12.9.zip](https://github.com/user-attachments/files/21867569/Kerbalism-Config_dev-9363_for_KSP1.8.0-1.12.9.zip)

~~[Kerbalism-Core_dev-9362_for_KSP1.8.0-1.12.9.zip](https://github.com/user-attachments/files/21856558/Kerbalism-Core_dev-9362_for_KSP1.8.0-1.12.9.zip)~~
~~[Kerbalism-Config_dev-9362_for_KSP1.8.0-1.12.9.zip](https://github.com/user-attachments/files/21856563/Kerbalism-Config_dev-9362_for_KSP1.8.0-1.12.9.zip)~~

~~[Kerbalism-Core_dev-9361_for_KSP1.8.0-1.12.9.zip](https://github.com/user-attachments/files/21841792/Kerbalism-Core_dev-9361_for_KSP1.8.0-1.12.9.zip)~~
~~[Kerbalism-Config_dev-9361_for_KSP1.8.0-1.12.9.zip](https://github.com/user-attachments/files/21841764/Kerbalism-Config_dev-9361_for_KSP1.8.0-1.12.9.zip)~~


gotmachine:
For reference, I ended up implementing the equalization mechanism requested by @KerballOne 

As a reminder, this additional PAW option is available either when an inflatable is being inflated, or when a deployable has just been deployed and is waiting for pressure to reach the pressure threshold. Enabling it will transfer atmosphere from the other enabled habs, which in some cases can help speed up reaching the enabled state, usually with the drawback that this will cause the vessel to loose the pressurized state/bonus.

Because of the caveats of that feature : 
- It isn't obvious that this will cause a loss of the pressurized state, potentially for days/weeks given how weak pressure controllers are.
- The actual speed-up will only be realized and significant if the volume of the enabled habs is several times larger than the volume of the inflatable, which won't be the case in most situations.
- If you didn't bring enough nitrogen to inflate/pressurize the inflatable/deployable, this won't change anything as you won't be able to reach the pressure threshold again. But you're now stuck in the the depressurized state, which wouldn't have happened if you didn't use the option.
- Those drawbacks / caveats aren't obvious and hard to assess, while what the average player will understand is that clicking that button is what you should do to deploy/inflate.

This is why additional heuristics drive the option availability, disabling it if we're reasonably sure that one of these practical issues will happen : 
https://github.com/Kerbalism/Kerbalism/blob/e77774580dc1506b19040a8bd911a160694934f7/src/Kerbalism/Modules/Habitat.cs#L107-L117

R-T-B:
I actually like and approve of @KerballOne's changes.

Got has said this PR is basically done, so I'll merge it now (though I am still testing it).  @KerballOne, if you want to implement the above, I see no personal issue with it.

src/Kerbalism/Database/VesselHabitatInfo.cs
```diff
@@ -71,6 +87,248 @@ internal void Save(ConfigNode node)
...
+
+				state = ObjectState.Unloaded;
+			}
+
+			if (state == ObjectState.Uninitialized)
+				return;
+
+			if (habitatWrappers.Count == 0)
+			{
+				if (vessel.isEVA)
```
KerballOne, line 143:
```
	// handle Kerbal being seated in external command seat
	if (modules[j] is KerbalSeat seat)
	{
		if (seat.Occupant.isKerbalEVA())
		{
			EVAStateUpdate(vessel, vd);
			if (state == ObjectState.Loaded && Features.Radiation)
				EVARadiationUpdate(vessel);
		}
	}
```
Same issue as https://discord.com/channels/418222516386004992/590004534051602452/1407378832255614996
But even less likely to occur.  Only if spawned with Kerbal in command seat.

Note:  Even if this bug is fixed, Kerbal will not spawn with resources to survive with Kerbalism Life Support needs.

PR #923, opened 2025-08-20
Backport of the Kerbalism 4 stock science recovery dialog integration

gotmachine:
On vessel recovery, the recovered science data will now show in the usual stock recovery UI instead of in a custom additional popup window. The code is backported from Kerbalism 4.

Not much testing done, seems to work without issues.


gotmachine:
This is purely a UI / cosmetic change, nothing else should be affected in any way.

PR #925, opened 2025-08-21
KerbalEVA bug fixes

KerballOne:
- Fixes bug that grants free Atmosphere on EVA when no Atmosphere is available from vessel.  
- Prevent bug for KerbalEVA taking Waste resulting in quick CO2 poisoning

### DIFF
--- a/GameData/KerbalismConfig/System/Habitat.cfg
+++ b/GameData/KerbalismConfig/System/Habitat.cfg
@@ -213,7 +213,7 @@
   RESOURCE
   {
 	name = Atmosphere
-	amount = 0		 // Kerbals suit up in a pressurized environment so they start pressurized.
+	amount = 0		 // Kerbal EVA suits must get atmosphere from the vessel habitat
 	maxAmount = 330	  // Liters
   }
 }
--- a/GameData/KerbalismConfig/System/Habitat.cfg
+++ b/GameData/KerbalismConfig/System/Habitat.cfg
@@ -213,7 +213,7 @@
   RESOURCE
   {
 	name = Atmosphere
-	amount = 330		 // Kerbals suit up in a pressurized environment so they start pressurized.
+	amount = 0		 // Kerbals suit up in a pressurized environment so they start pressurized.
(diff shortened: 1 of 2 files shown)


GameData/KerbalismConfig/System/Habitat.cfg
```diff
@@ -213,7 +213,7 @@
   RESOURCE
   {
 	name = Atmosphere
-	amount = 330		 // Kerbals suit up in a pressurized environment so they start pressurized.
+	amount = 0		 // Kerbals suit up in a pressurized environment so they start pressurized.
```
gotmachine, line 216: The comment is now wrong, you should change it to something like
// Atmosphere will be transferred upon EVAing

Also, this change will be an issue if an EVA Kerbal is spawned without EVAing, this should be verified. I can think of at least one potential case : launching in an external command seat. Theoretically, a mod could also be doing this, but right now I can't think of any that does.

src/Kerbalism/System/Callbacks.cs
```diff
@@ -198,6 +198,12 @@ void ToEVA(GameEvents.FromToAction<Part, Part> data)
 					continue;
 				}
 
+				// do not take out waste
```
gotmachine, line 201: This isn't the right fix.
The problem is on the next line (207), the generic formula doesn't make sense for Atmosphere and WasteAtmosphere, we should instead account for the current level on the vessel :
```cs
double quantity = 0.0;

// eva prop is handled differently
if (res.resourceName == evaPropName)
	continue;
// conserve pressure / CO2 level
else if (res.resourceName == Habitat.AtmoResName || res.resourceName == Habitat.WasteAtmoResName)
	quantity = resources.GetResource(data.from.vessel, res.resourceName).Level * res.maxAmount;
// for other consumables (food, water, oxygen...), try to fill it but don't take more than the kerbal share amongst the total crew
else
	quantity = Math.Min(resources.GetResource(data.from.vessel, res.resourceName).Amount / tot_crew, res.maxAmount);

if (quantity == 0.0)
	continue;
```

Issue #930, opened 2025-08-23
[Feature Request] Add minimum (and maybe maximum) comm link speed as an experiment requirement.


njits23:
I'm currently working on a Kerbalism experiment where the goal is to have the player transmit live TV on the way to the moon, and whilst landed on the moon. However, I'm having difficulty enforcing a sufficient comm speed to transmit the data live. 
Kerbalism is obviously aware of what the current comm link speed of a vessel is, as it's listed in the vessel details page. However, this information is not usable as an experiment requirement. I would greatly appreciate it if the ability to specify a minimum commlink speed to run an experiment was made available.

I took a look at https://github.com/Kerbalism/Kerbalism/blob/master/src/Kerbalism/Database/VesselData.cs and https://github.com/Kerbalism/Kerbalism/blob/master/src/Kerbalism/Science/ExperimentRequirements.cs and it seems like it'd be possible to implement, but I don't know which vessel field to reference.

PR #931, opened 2025-08-23
Add min/max transmission rate experiment requirement

njits23:
This PR adds the min/max transmission rate experiment requirement I requested in https://github.com/Kerbalism/Kerbalism/issues/930 

Once merged, the wiki page for experiments should be updated to include this new requirement: https://github.com/Kerbalism/Kerbalism/wiki/TechGuide-~-PartModules-~-Experiment


njits23:
Finally got Kerbalism to generate localisations for all languages by installing it (and its dependencies) into an otherwise completely bare install.

gotmachine:
>  There does appear to be a problem in the localisation generator where it'll replace //UNTRANSLATED lines with the English translation when the localisation is regenerated, even if there isn't an actual translation.

Yeah I noticed that too, not too big of an issue.

PR looks good to me, I think this can be merged.

PR #935, opened 2025-08-24
EVA Kerbal handling refactor

gotmachine:
Following the habitat/pressure refactor, this kinda finish the job by refactoring how EVA Kerbals are handled :

### Changes

- EVA kerbals now have a Habitat module set to an inescapable `evaKerbal` state, which is handling all the LS related stuff on them. All the scattered bits handling them has been consolidated as much as possible, and various places with special handling have been simplified. Overall, this make the codebase a little bit cleaner, and was necessary to properly support kerbals in external command seats.
- Fixed incorrect handling of EVA suit lamps / lights
- Fixed incorrect freezing of EVA animation on dead kerbals
- Dead EVA kerbals now can't interact with anything through the correct control lock, which is more reliable and simpler than what was done previously.
- Implemented support for EVA kerbals in external command seats, including launching in an ECS.
- Fixed support of EVA construction. This wasn't handled correctly, resulting in VesselData/PartData/HabitatData becoming out of sync, triggering various issues and exceptions.
- Some fixes to various order of initialization and synchronization related issues, notably causing the profile (rules, processes) to be evaluated while the VesselData wasn't yet ready (typically on scene start for the first 2-3 fixedupdates). This still probably has some holes, the overall lifecycle is quite messy.
- Various small bugfixes here and there.

### Config changes
The way EVA kerbals are now handled is by adding an habitat module to them, set to the appropriate state, this should look like this :
```
MODULE
{
  name = Habitat
  state = evaKerbal
  volume = 0.33 // m^3, match the default profile Atmosphere resource capacity of 330 that was used in previous versions
  surface = 1.5 // m^2, only has an effect for kerbals seated in an external seat, as they are technically a part of the vessel
}
```
The `Atmosphere` and `WasteAtmosphere` resources should **not** be added with a MM patch anymore, but won't conflict if present.
`on_eva` supplies will still work as they did previously. 

### Backward compatibility

To ensure backward compatibility with third party profiles, the habitat module will be dynamically added on EVA kerbal parts if not present. Its volume will be set to the capacity of the `Atmosphere` resource divided by 100 if that resource is defined (it should be), or will default to 3.3 m^3 otherwise. The surface will default to 1.5 m^2.

Behavior for saves having EVA kerbals in them *should* be seamless.

### QA

I have done a fair amount of testing, but some of the core systems changes could potentially have side effects I didn't anticipate. If anybody is willing to test things further, this can't hurt.

Here is a dev build with the changes : 
[Kerbalism-Core_dev-9368_for_KSP1.8.0-1.12.9.zip](https://github.com/user-attachments/files/21963806/Kerbalism-Core_dev-9368_for_KSP1.8.0-1.12.9.zip)
[Kerbalism-Config_dev-9368_for_KSP1.8.0-1.12.9.zip](https://github.com/user-attachments/files/21963820/Kerbalism-Config_dev-9368_for_KSP1.8.0-1.12.9.zip)

~~[Kerbalism-Core_dev-9367_for_KSP1.8.0-1.12.9.zip](https://github.com/user-attachments/files/21958839/Kerbalism-Core_dev-9367_for_KSP1.8.0-1.12.9.zip)~~
~~[Kerbalism-Config_dev-9367_for_KSP1.8.0-1.12.9.zip](https://github.com/user-attachments/files/21958841/Kerbalism-Config_dev-9367_for_KSP1.8.0-1.12.9.zip)~~


KerballOne:
Resulting Atmo/WasteAtmo is 3300, not 330.
And with the extra volume, EC only lasts 16 minutes with Climatization consuming -0.035/s
image

PR #937, opened 2025-08-31
Fix signal notification checkbox not showing up in vessel config

### LINKED ISSUE #823: RemoteTech not working properly with Kerbalism 3.16
Kerbalism Version - 3.16
KSP Version - Latest from Steam
Mods - Admittedly, I'm running several mods beyond Kerbalism/RemoteTech. 
Remotetech Version - 1.9.12
[Playerlog.zip](https://github.com/Kerbalism/Kerbalism/files/9377314/Playerlog.zip)

Issue - Per the attached screenshot, the check box to stop loss of signal connection from aborting auto warp is not present at all. My understanding is that the checkbox for signal loss should be under "messages."

screenshot1

Begus001:
This is a workaround for RemoteTech users to always show the checkbox to toggle signal lost/regained notifications, because it does not show up when using RemoteTech, which, if signal notifications are enabled in the global Kerbalism settings, leads to annoying notifications (and timewarp cancelling) for every craft that cannot be disabled without editing the save file.

I am not sure why RemoteTech does not show up as a comm handler in `API.Comm.handlers` though.

Fixes #823.

### DIFF
--- a/src/Kerbalism/UI/VesselConfig.cs
+++ b/src/Kerbalism/UI/VesselConfig.cs
@@ -53,8 +53,7 @@ namespace KERBALISM
 				p.AddContent(Local.VESSELCONFIG_supply, string.Empty, tooltip);//"supply"
 				p.AddRightIcon(vd.cfg_supply ? Textures.toggle_green : Textures.toggle_red, tooltip, () => p.Toggle(ref vd.cfg_supply));
 			}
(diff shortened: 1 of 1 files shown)


Aebestach:
I believe this revised PR should resolve #823.
@R-T-B Should be good to merge now.

R-T-B:
Yep, looks good.  We should put out a new preview dev build soon as well.

Issue #942, opened 2025-09-02
Kerbals assigned to a sickbay never loose the effect after recovering the vessel


gotmachine:
Assigning a Kerbal to a sickbay "cure" (or just having the Kerbal in the part for sickbays with `cureEverybody = true`) will add the cure to the kerbal.
The module continuously check if the Kerbal is still in the part, so the cure will be removed if the Kerbal is transferred or goes on EVA.

However, it will never be removed when the vessel is recovered, resulting in cures being applied out of nowhere on such recovered Kerbals if they are on a vessel where a matching sickbay is active (even if they aren't in the sickbay part / aren't affected to a sickbay slot).

Aebestach:
#948 has been fixed, and this issue will now be closed.

Issue #941, opened 2025-09-02
Repairing doesn't apply until vessel is reloaded


gotmachine:
Some modules don't immediately leave the "broken" state when repairing them.
You have to reload the vessel (for example by quicksaving/quickloading) for the module to become functional again.

This notably affect ProcessControllers, but also Experiments (not much impact for those as the default config doesn't have reliability on experiments).
I think the stock ModuleScienceExperiment and ModuleLight also have a similar issue

The reason is the `Reliability.Apply()` method not handling the "repair" event at all for those modules (`b == false`), only the "break" event (`b == true`) : 
https://github.com/Kerbalism/Kerbalism/blob/7fecbbb797f37a9369427c627a489bd15f803fce/src/Kerbalism/Modules/Reliability.cs#L882-L889
https://github.com/Kerbalism/Kerbalism/blob/7fecbbb797f37a9369427c627a489bd15f803fce/src/Kerbalism/Modules/Reliability.cs#L942-L949
https://github.com/Kerbalism/Kerbalism/blob/7fecbbb797f37a9369427c627a489bd15f803fce/src/Kerbalism/Modules/Reliability.cs#L932-L939
https://github.com/Kerbalism/Kerbalism/blob/7fecbbb797f37a9369427c627a489bd15f803fce/src/Kerbalism/Modules/Reliability.cs#L899-L914

However, fixing this might be tricky and cause side effects, because when repairing, the state of ProcessControllers and Experiments is actually handled by the Configure module if those are part of configure setup : 
https://github.com/Kerbalism/Kerbalism/blob/7fecbbb797f37a9369427c627a489bd15f803fce/src/Kerbalism/Modules/Reliability.cs#L598-L601

So calling `ReliabilityEvent()` for those modules when repairing them will mess with the intended state. It might be possible to at least fix the issue for configure-handled modules by still calling it but only lifting the broken flag/state. 

Alternatively, maybe the right fix is to fully handle both state changes from `Reliability.Apply()` but to call `Configure.DoConfigure()` after instead of before. Not entirely sure what the other side effects could be.

Aebestach:
#948 has been fixed, and this issue will now be closed.

Issue #940, opened 2025-09-02
Self-consuming processes (RTG, Non-regen scrubbers...) don't work correctly


gotmachine:
Self-consuming processes are processes that have their own pseudo-resource as an input.
This is used to represent processes that have a limited lifetime and whose efficiency decrease over time.
For example, the RTG process is consuming its own `_RTG` resource : 
```
Process
{
  name = radioisotope generator
  title = #KERBALISM_Process_RadioisotopeGenerator
  modifier = _RTG
  input = _RTG@0.000000001888 // 28.8 kerbin-year half-life
  output = ElectricCharge@1.0
}
```
Other such processes are the non-regen scrubber, possibly others in support / third party configs.

This is buggy when the process is controlled by a `ProcessController`.
Specifically, the pseudo-resource amount will be refilled every time `ProcessController.Start()` is called, so every time the vessel is re-loaded, negating the intended effect.

This is because of the reliability integration which forcibly reset the pseudo-resource amount to its configured value based on the broken / not broken state :
https://github.com/Kerbalism/Kerbalism/blob/7fecbbb797f37a9369427c627a489bd15f803fce/src/Kerbalism/Modules/ProcessController.cs#L111

Not sure since when this is happening, and not sure how this can be fixed either given the spaghetti plate involved.

R-T-B:
I'm just spitballing but could a simple fix like this work?  I'll test if it has any potential.

`		/// <summary> Set the enabled/disabled state of a process
		/// <para> Use the process_capacity parameter to set the pseudo resource amount for the process,
		/// an amount of 0.0 disables the process, any non-zero value is a multiplier of the process.
		/// </para> </summary>
		public static void SetProcessEnabledDisabled(Part p, string res_name, bool enable, double process_capacity)
		{
			bool resNotExist = !p.Resources.Contains(res_name);
			if (resNotExist)
			{
				Lib.AddResource(p, res_name, 0.0, process_capacity);
			}

			if (enable && resNotExist)
			{
				Lib.SetResource(p, res_name, process_capacity, process_capacity);
				return;
			}

			if (enable && !resNotExist)
			{
				Lib.SetResource(p, res_name, p.Resources[res_name].amount, process_capacity);
				return;
			}

			else
			{
				// Never remove the resource capacity, otherwise checks against
				// the pseudo resource might fail
				Lib.SetResource(p, res_name, 0.0, process_capacity);
			}
		}`

R-T-B:
[...]

Issue #642, opened 2020-05-30
Setting the same process on two different chem plant slots will not work


basilicofresco:
Hi! I noticed that when I set the same process on more than one slot of a chemical plant, then only the first one will actually work. The problem is visible on the planner and on the field. It affects all the converters (Chemical plant, Convert-O-Tron 125, Convert-O-Tron 250) and I guess all the processes. I just tried with water electrolysis and waste incinerator.
![kerbalism2](https://user-images.githubusercontent.com/31368421/83335233-97e9a300-a2ab-11ea-8457-e0905146c9ce.png)

Could the problem be related to the fact that there is just one switch on the pop-up window of the converter?
![kerbalism3](https://user-images.githubusercontent.com/31368421/83335720-f95f4100-a2ae-11ea-8349-86512fe8dabc.jpg)

I tested it with the usual clean installation with just KSP 1.9.1.2788 ("Science = false" in settings.cfg) + Breaking Ground 1.4.1 + Kerbalism 3.9. The problem is present both in career and science mode.
Thanks!

gotmachine:
Yep, known issue. Won't fix as the configure module is going away in next major version.

gotmachine:
To clarify a bit, I haven't investigated in detail, but due to how the configure module is implemented, I don't see how this can be fixed without a major refactor.

Aebestach:
#948 has been fixed, and this issue will now be closed.

PR #948, opened 2025-09-12
ProcessController and related code refactor

gotmachine:
This notably aim at fixing the following issues :
- #642
- #940
- #941 
- #942 

Main change is reverting this commit : ac24298f56a7111e76d561c673e916dc840ee433 (note that other related changes were made following this commit as it broke a bunch of stuff)

Originally, ProcessControllers enabled state toggling (either by the user or by a Reliability module) was done by toggling the pseudo-resource `flowState`. The mentioned commit changed that logic to instead alter the pseudo-resource amount (setting it either zero or maxAmount), which notably broke self-consuming processes (see issue #940), and more generally made things a lot messier.

Aside from fixing issue #940, reverting this change was necessary to implement the fix for issue #642, and more generally it make the overall implementation a lot cleaner.

The PR implement a save/craft persisted data migration code to preserve backward compatibility with existing saves from 3.24 and below. However, do note that once migrated to 3.25 or above, the save will not work correctly anymore in 3.24 and below, so backing up saves is recommended if for any reason people want to downgrade. This being said, the issues should be relatively minor, mainly user-disabled or reliability-broken PCs being forever stuck in the disabled state.

The PR implement a proper fix for issue #642 by allowing the `IConfigurable.Configure()` method to specify a multiplier for when multiple slots are assigned to the same setup. The ProcessController module then implement proper handling of the multiplier by applying it to the pseudo-resource amount/maxAmount in its `IConfigurable.Configure()` implementation.

Due to the changes being somehwat risky, this should probably be playtested a bit before merging/releasing. 

Dev builds (remember to backup your save as it won't be compatible with 3.24 anymore) :
[Kerbalism-Core_dev-9386_for_KSP1.8.0-1.12.9.zip](https://github.com/user-attachments/files/22296982/Kerbalism-Core_dev-9386_for_KSP1.8.0-1.12.9.zip)
[Kerbalism-Config_dev-9386_for_KSP1.8.0-1.12.9.zip](https://github.com/user-attachments/files/22296957/Kerbalism-Config_dev-9386_for_KSP1.8.0-1.12.9.zip)

### DIFF
--- a/GameData/Kerbalism/Kerbalism.version
+++ b/GameData/Kerbalism/Kerbalism.version
@@ -3,7 +3,7 @@
   "URL": "https://raw.githubusercontent.com/Kerbalism/Kerbalism/master/GameData/Kerbalism/Kerbalism.version",
   "DOWNLOAD": "https://github.com/Kerbalism/Kerbalism/releases",
   "CHANGE_LOG_URL": "https://github.com/Kerbalism/Kerbalism/blob/master/CHANGELOG.md",
-  "VERSION": {"MAJOR": 3, "MINOR": 24, "PATCH": 0, "BUILD": 0},
+  "VERSION": {"MAJOR": 3, "MINOR": 25, "PATCH": 0, "BUILD": 0},
   "KSP_VERSION": {"MAJOR": 1, "MINOR": 12, "PATCH": 0},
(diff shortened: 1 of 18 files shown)


R-T-B:
This has passed testing for several days of gameplay so I am going to merge it.

Issue #949, opened 2025-09-17
When loading craft from different save, unresearched experiments cannot be removed from avionics


byt3c:
The issue is related to this thread in RP-1
[https://github.com/KSP-RO/RP-1/issues/2561#issuecomment-3304301595](https://github.com/KSP-RO/RP-1/issues/2561#issuecomment-3304301595)

Basically loading a craft that has unresearched experiments in avionics, makes it impossible to remove from experiments menu.

Clayell:
<img width="811" height="478" alt="Image" src="https://github.com/user-attachments/assets/21e9bddf-898a-42d8-a7b1-f06643870524" />

I'll post your image here since it seems you forgot. Notice how "Video Imaging" cannot be seen in the menu.

Clayell:
Should this be closed?

R-T-B:
Probably?  I'll close for now but feel free to reopen if this is still happening.

byt3c:
i haven't seen that happening, but when technology is being researched sometimes it appears as selected- i'm not sure if that blocks the slot or not.

PR #951, opened 2025-09-19
Fix for issue #949

gotmachine:
Did zero testing for side effects, this should work alright but one never knows with the configure module...


R-T-B:
I'll start testing this.  I plan to push a release today, so this will probably be in next release, but will merge as soon as the release is properly tested and confirmed to behave as expected.

PR #956, opened 2025-10-10
proposal: produce food constantly from greenhouses, remove harvests

jessicest:
proposal: let's make greenhouses deliver food constantly.

Why: greenhouses interfere with timewarp, to have the player do a chore that involves no actual decisions.

If you have a mining base or space station with many greenhouses, and you try running 10,000x timewarp, you'll need to interrupt your timewarp every 7-8 minutes to click "harvest" on each one, or your base will have a loss of food/oxygen production.

For gameplay purposes, let's just allow them to grow food constantly.

This PR is a demo implementation for testing and feedback:

 * it retains the food rates of kerbalism and sspx greenhouses
 * it stalls out harvests so they don't disrupt the food production

Ideally if this proposal gets strong support, this PR should include some code changes as it retains these flaws:

 * the "harvest" is still part of the UI; it's vestigial with this change
 * food growth seems to continue when co2 is depleted
 * it doesn't help players who play with other profiles, like Realism Overhaul players -- who could also use a break :D


Aebestach:
Thanks for the proposal, but this approach is only a temporary fix and doesn't fully address the issue. 
I'll close this PR; you can refer to #1073 for more details.
Thank you once again.

PR #958, opened 2025-10-17
Add API methods to get average rates of vessel resources

Phantomical:
I have made these match the existing API methods. I _think_ `AverageRate` is the right field to expose to users. I'm happy to adjust that if that is not the case.


R-T-B:
This looks good to me.  I'll merge it.  Sorry for missing this for so long.

PR #963, opened 2025-10-24
Update ReStockPlus_Science.cfg & Add Squad_Science.cfg

Aebestach:
| ReStockPlus_Science | Squad_Science |
|--------|--------|
| Snipaste_2025-10-24_13-19-26 | Snipaste_2025-10-24_13-58-19 | 

# ReStockPlus
https://github.com/Kerbalism/Kerbalism/blob/ef685c06b665eccfd5d8e79a108c0f8dca7381de/GameData/KerbalismConfig/Support/ReStockPlus_Science.cfg#L80

The `Experiment Return Unit` should be rebalanced. According to the title and description of this part, it should be able to store more data, rather than the original minimum of 0.5MB and a maximum of 2MB. This is far from enough for the large amount of data that needs to be returned, so I adjusted the current capacity.

```
@PART[restock-sciencebox-stack-1]:NEEDS[RestockPlus,FeatureScience]:AFTER[KerbalismDefault]
{
	@mass = 0.5
	@cost = 2500
	%MODULE[HardDrive]
	{
		%dataCapacity = 500
		%sampleCapacity = 4
	}
}
```
The adjustments are as follows:
1. Increase the mass to 0.5t
2. Minimum storage capacity to 500MB, maximum to 2GB
3. Cost increased to 2500

# Squad_Science
```
@PART[ScienceBox]:NEEDS[FeatureScience]:FOR[KerbalismDefault]
{
	@mass = 0.35
	@cost = 1500
	@MODULE[HardDrive]
	{
		@dataCapacity = 250
		@sampleCapacity = 8
	}
}
```

Add similar configuration for `Experiment Storage Unit`. 
However, I found that the original configuration seems to have set the capacity to 0. I wonder why this is considered? If it is a data return unit, it should also have a capacity?

### DIFF
--- a/GameData/KerbalismConfig/Support/Squad_Science.cfg
+++ b/GameData/KerbalismConfig/Support/Squad_Science.cfg
@@ -1,0 +1,10 @@
+@PART[ScienceBox]:NEEDS[FeatureScience]:FOR[KerbalismDefault]
+{
+	@mass = 0.35
+	@cost = 1500
+	@MODULE[HardDrive]
+	{
+		@dataCapacity = 250
+		@sampleCapacity = 8
+	}
+}


R-T-B:
I can't help but agree with these changes, and as there are just config changes reallly, they are a safe merge.  Doing so.

PR #972, opened 2025-12-14
add datarate information to CONTROL PATH window

### LINKED ISSUE #969: Signal `CONTROL PATH` should show the b/s at each hop
I had impossibly high connection speeds, so I thought Kerbalism only ever considered the speed of the active vessel to limit upload speed.

This is infact false, Kerbalism properly finds the bottleneck (it doesn't consider congestion\*) in the path.

\*we would need big changes to make congestion fun.

However this UI is so ~~useless~~ in need in love I didn't knew it worked properly:
Image

The `%` were meaningless to me,
I believe it should be `b/s` (humanized `kb/s`, ...) either alone with `%` on hover or next to the `%`.

Would be helpful for searching bottlenecks too, a really fast connection with a slow `%` is still faster than a really slow connection with good `%`.

Jorropo:
Fixes #969

- Added datarate to CONTROL PATH window
- Colorize the datarate and signal strength based on the signal strength
- move strength to hover (it's already implied by color)
- unified across commnet implementations (semi-refactor)

### Before

image

### After

image

### DIFF
--- a/src/Kerbalism/Comms/CommHandlerCommNetBase.cs
+++ b/src/Kerbalism/Comms/CommHandlerCommNetBase.cs
@@ -25,8 +25,11 @@ namespace KERBALISM
 			{
 				connection.strength = 0.0;
 				connection.rate = 0.0;
+				connection.hop_datarate = 0.0;
 				connection.target_name = string.Empty;
-				connection.control_path.Clear();
+				connection.next_hop = Guid.Empty;
+				connection.hop_distance = 0.0;
+				connection.hop_max_distance = 0.0;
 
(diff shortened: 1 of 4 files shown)


Jorropo:
Fixed the logic such that it shows the kb/s at each hop (it used to show the min along the path):
image

R-T-B:
I'm 99% sure you don't need to do anything just like you say, I just want some actual beta testing proving that concept as one of the RO devs expressed mild concern that it COULD happen.  As soon as we get some live testing results that meet with their approval I'll most likely merge this.

Issue #974, opened 2025-12-17
radio path does not update with RT when switching from relay → KSC


Jorropo:
See this situation:

<img width="620" height="538" alt="Image" src="https://github.com/user-attachments/assets/9b1246c0-d998-4ae5-b4dc-8e16e8fdc57b" />

KSC just rose over Kerbin's horizon, but it thinks I am connected through `KommSat 1.0 C` since it were it's last hop.
It also think the datarate is terrible when it is not when `KommSat 1.0 C` was actually the relay (it has an Active Vessel antenna with 200Gm of range and I am using the Root model).

Aebestach:
fixed by #975

PR #975, opened 2025-12-18
with RT update path's next_hop when switching from vessel → KSC radio link

Jorropo:
Updates #974

In fact 947 is in fact two different bugs:
1. the radio path sometimes isn't updated affecting b/s and CONTROL PATH
2. the radio path sometimes isn't updated affecting CONTROL PATH

This fix the second of the two which I've introduced in f08fa1601de4dd16e679f08be2140d30837df4f3, I am pretty sure the first one was already there since I didn't changed the rate calculation. But the first bug is also much rarer and I don't know what triggers it.


R-T-B:
This seems sensible.  Merging for final testing before hopefully including with tomorrows release...

Issue #981, opened 2026-01-07
Adding OKTO2 to vessel breaks simulation panel


ShadowSTARS616:
When adding the stock OKTO2 to a vessel in the VAB/SPH, Kerbalism's simulation panel in the bottom-right corner loses all displayed data. Removing the part restores normal functionality. 
This issue occurs specifically with the OKTO2 probe core and does not happen with any other unmanned probe cores or manned command pods.

before:
![Image](https://github.com/user-attachments/assets/ff15d7f3-34bc-434f-9409-18be1fca05e1)
adding OKTO2:
![Image](https://github.com/user-attachments/assets/f1a64b55-e6a7-40ca-9361-687bddf844f4)

When the OKTO2 is added, the following errors appear in KSP.log:

[LOG 12:31:26.927] probeCoreOcto2.v2 added to ship - part count: 125
[EXC 12:31:27.007] NullReferenceException: Object reference not set to an instance of an object
	KERBALISM.Lib.CrewCount (Vessel v) (at <0d51f7fc071d45809eb1d485b198894d>:0)
	KERBALISM.Comforts..ctor (System.Collections.Generic.List`1[T] parts, System.Boolean env_firm_ground, System.Boolean env_not_alone, System.Boolean env_call_home) (at <0d51f7fc071d45809eb1d485b198894d>:0)
	KERBALISM.Planner.VesselAnalyzer.Analyze_qol (System.Collections.Generic.List`1[T] parts, KERBALISM.Planner.ResourceSimulator sim, KERBALISM.Planner.EnvironmentAnalyzer env) (at <0d51f7fc071d45809eb1d485b198894d>:0)
	KERBALISM.Planner.VesselAnalyzer.Analyze (System.Collections.Generic.List`1[T] parts, KERBALISM.Planner.ResourceSimulator sim, KERBALISM.Planner.EnvironmentAnalyzer env) (at <0d51f7fc071d45809eb1d485b198894d>:0)
	KERBALISM.Planner.Planner.Update () (at <0d51f7fc071d45809eb1d485b198894d>:0)
	KERBALISM.Launcher.Update () (at <0d51f7fc071d45809eb1d485b198894d>:0)
	KERBALISM.UI.Update (System.Boolean show_window) (at <0d51f7fc071d45809eb1d485b198894d>:0)
	KERBALISM.Kerbalism.Update () (at <0d51f7fc071d45809eb1d485b198894d>:0)
[...]

PR #982, opened 2026-01-07
Fix NullReferenceException and Kerbalism UI errors for Probodobody OKTO2 part in VAB/SPH (Comfort.cfg update)

Aebestach:
Update `Comfort.cfg` to fix issues with NullReferenceException and Kerbalism UI exceptions when using the **Probodobody OKTO2** part in VAB/SPH. See #981 for details.


ujihisa:
:+1:
I'd like to use okto2 in Kerbalism

PR #985, opened 2026-01-10
Update NFElectric.cfg for renamed and new fission reactors

Sandoleathy:
This PR updates `NFElectric.cfg` to be compatible with the current version of Near Future Electric.

Specifically, the fission reactor definitions in NFE have changed: some reactor names were renamed and several new reactor types were introduced. As a result, Kerbalism could not correctly handle the background power generation of these reactors.

This change updates the configuration so that Kerbalism correctly recognizes the current fission reactors and applies the proper background power generation values.


LouisB3:
I believe the recycler name also needs to addressed (missing a "-1" at the end)

Sandoleathy:
Yes! Thanks for your suggestion. I will fix it later

Sandoleathy:
Based on the comment, the incorrect reactor name `nfe-nuclear-recycler-25-1` has been fixed.

In addition, this PR adds support configurations for the PlanetSideExplorationTechnologies mod.  
Specifically:
- Added habitat volume for the inflatable base part `Benjee10_base_HDU_attic`.
- Increased hard drive storage and sample capacity for the command module `Benjee10_MMSEV_baseHDU`.

I'm very sorry that I didn't place the support files for PanetSideExplorationTechnologies in the new branch. I didn't realize that all the contents in the same branch would be mapped to this PR.

Sandoleathy:
Deleted files relevant with the mod PlanetSideExplorationTechnologies and they are moved to pull request #989

Issue #960, opened 2025-10-23
Potential issue in Kerbalism multi-star detection causes solar panels to stop working


Aebestach:
| NonKerbalism | Kerbalism |
|--------|--------|
| <img width="696" height="554" alt="Image" src="https://github.com/user-attachments/assets/b60d1387-f1b5-475a-a66f-6a66dae72598" /> | <img width="718" height="465" alt="Image" src="https://github.com/user-attachments/assets/32395fd2-9628-4564-8664-2208e9283f7e" /> |
| <img width="667" height="458" alt="Image" src="https://github.com/user-attachments/assets/0b9d7832-0e55-4bd4-95cb-e6f5cec13592" /> | <img width="654" height="460" alt="Image" src="https://github.com/user-attachments/assets/93f24de2-d0da-40f5-8a12-ba6be6c832e7" /> | 

https://github.com/Kerbalism/Kerbalism/blob/ef685c06b665eccfd5d8e79a108c0f8dca7381de/src/Kerbalism/Modules/SolarPanelFixer.cs#L1-L1801

From [Kopernicus v209_RC1](https://github.com/Kopernicus/Kopernicus/releases/tag/release-209_RC1) to [Kopernicus v212](https://github.com/Kopernicus/Kopernicus/releases/tag/release-212). @R-T-B and I have implemented a more robust version of the solar panel module in Kopernicus, originally based on Kerbalism’s design. So far, no abnormal behavior has been observed in these versions.

As shown in the attached image, Kerbalism still has issues with multi-star system detection, which prevents solar panels from functioning correctly. I believe it’s time for this issue to be addressed, and @R-T-B assistance might be needed for this fix.

Aebestach:
Electricity is still being produced, but it's not showing up in PAW; this bug is currently being fixed. Other issues are also being addressed.

Aebestach:
Fixed in v3.40

PR #986, opened 2026-01-18
Fix Multi-Star Solar Output Issues and Add UniversalStorage2 Support

Aebestach:
# Bug Fixes and New Features in `SolarPanelFixer.cs`

This PR primarily aims to fix and improve the issues described in **#960**, while significantly enhancing multi-star behavior, analytic power calculation, UI clarity, and background simulation consistency.

---

### 1. Enhanced PAW UI

* Reworked the PAW display to **split status information into multiple dedicated fields**, instead of packing everything into a single `panelStatus` line.
* Newly separated fields include:

  * **Solar Panel Mode**
  * **Energy Output**
  * **Exposure**
  * **Wear**
* Improves readability and reduces ambiguity in panel state reporting.

---

### 2. Higher Output Display Precision

* Dynamically increases decimal precision for **very small (but non-zero)** power outputs.
* Prevents low power generation from being incorrectly shown as **“0”**.

---

### 3. Improved Multi-Star Auto-Tracking

Auto mode no longer blindly tracks `EnvMainSun`.

* **Analytic mode**
  → Selects the star with the **highest `SolarFlux`**.

* **Non-analytic mode**
  → Selects the **brightest visible** star (`SunlightFactor > 0.05`).

* **Fallback behavior**

  * If *all* stars are blocked, the panel tracks the **geometrically nearest** star.
  * This keeps the panel pre-aligned when exiting shadow, reducing tracking lag.

---

### 4. Reworked Multi-Star / Analytic Power Model

Introduces new analytic power computation paths:

* `CalculateMultiStarPowerAnalytic()`
* `CalculateLandedMultiStarPower()`

Key characteristics:

* Accumulates **per-star power contributions**
* Accounts for:

  * Panel type (**flat / cylindrical / spherical**)
  * Tracking vs non-tracking behavior
* **Tracking panels**:

  * Full efficiency for the *tracked main star*
  * Reduced contribution from secondary stars using **directional cosine weighting**

---

### 5. Analytic Eclipse & Daylight Duty-Cycle Simulation

* **Landed vessels**:

  * Daylight fraction is computed analytically from **latitude / declination**
  * Special handling for **tidally-locked bodies**
* Improves accuracy compared to previous time-averaged approaches.

---

### 6. More Consistent Background (Unloaded) Behavior

* `BackgroundUpdate()` now **recomputes multi-star power explicitly**, instead of relying on
  `persistentFactor × EnvSolarFluxTotal`.
* This fixes inconsistencies at **high timewarp / unloaded vessels**.

**Physical correctness improvement:**

* Old logic could still count flux from stars *behind* the panel.
* New logic clamps secondary-star contribution using **cosine directionality**:

  * No power contribution when the star is behind the panel.

---

### 7. Smoother Multi-Star Status Transitions

* In multi-star systems:

  * If the *tracked star* is shadowed or poorly oriented,
  * but **total output remains significant** due to other stars,
* The panel is treated as effectively **“Exposed”**.
* This reduces jittery state transitions in the UI and Kerbalism panel.

---

### 8. Multiple Resource Support

* Added support for solar panel modules that:

  * Use `ModuleDeployableSolarPanel`
  * **Do not generate `ElectricCharge`**
* Improves compatibility with custom and modded resource setups.

---

### 9. UniversalStorage2 Support

* Added explicit compatibility support for **UniversalStorage2** solar panel modules.

---

## 🐞 Bug Fixes

* Fixed a UI bug where an incorrect **“in shadow”** error could appear when updating solar panel status in the Kerbalism panel.
* Removed all logic and UI references related to `persistentFact`, including **average power factor**, since it is no longer used.
* Fixed a **background resource processing bug** where:

  * Solar panels continued generating power even when reliability had degraded to **faulty or worse**.

---

## ⚠️ Known Issues

1. **Players must reload all already-launched vessels** for updated configurations to take effect.

---

## 🧪 Testing Required

Given the scope and complexity of this PR, **extensive testing is required**:

1. **RSS environments** (RO / RP-1)
2. Additional **multi-star mods**, validating:

   * Real-time mode
   * Analysis mode
   * Landed / splashed / in-flight behavior
3. More comprehensive testing in the **stock environment**
4. Validation on both **old and new save archives**
5. Performance testing to ensure no regressions under:

   * High timewarp
   * Large vessel counts
   * Multi-star systems

> ⚠️ While some extreme-case testing has already been performed, more volunteers and broader coverage are strongly recommended.

---


R-T-B:
After some discussion, we have decided this will be mainlined, but it will come after the next bugfix release.  This is because of the one bug you note, where players have to reload their existing in-flight craft.  For players with a large career and lots of craft that may be upsetting, but we plan to do this in sync with next release and mark it as a "major feature" release thus making it obvious that there may be changes, and reducing the surprise.  Also this allows existing users who would be bothered by this change to get other bugfixes.

Sorry it's taken so long to get to this, and you are now on schedule for merge in the next month at worst case.  Thank you for your contribution!

PR #989, opened 2026-01-20
Add support file for mod PlanetSideExplorationTechnologies

Sandoleathy:
- Added habitat volume for the inflatable base part Benjee10_base_HDU_attic.
- Increased hard drive storage and sample capacity for the command module Benjee10_MMSEV_baseHDU.
- Add electric generation for all **  wind turbine generator **. The electricity production in the background is a fixed value. This value is the same as the one shown in the original tab of the wind turbine.

### DIFF
--- a/GameData/KerbalismConfig/Support/PlanetSideExplorationTechnologies.cfg
+++ b/GameData/KerbalismConfig/Support/PlanetSideExplorationTechnologies.cfg
@@ -1,0 +1,101 @@
+@PART[Benjee10_base_HDU_attic]:NEEDS[Kerbalism,PlanetSideExplorationTechnologies]:AFTER[KerbalismDefault]
+{
+    MODULE
+    {
+        name = Habitat
+        inflate = true
+        deployed = false
+        volume = 55
+        surface = 65
+    }
+}
+
+@PART[Benjee10_MMSEV_baseHDU]:NEEDS[Kerbalism,PlanetSideExplorationTechnologies]:AFTER[KerbalismDefault]
+{
+    @MODULE[HardDrive]
+    {
+        @dataCapacity = 4096
+        @sampleCapacity = 128
+    }
+}
+
+// Turbines
+@PART[Benjee10_MMSEV_baseTurbineGiant]:NEEDS[Kerbalism,PlanetSideExplorationTechnologies]:AFTER[KerbalismDefault]
+{
+    !MODULE[ModuleGenerator] {}
+    MODULE
+    {
+        name = ProcessController
+        resource = _ECGen
+        title = Turbine
+        capacity =50.0
+        running = false
+        toggle = true
+    }
+}
+@PART[Benjee10_MMSEV_baseTurbineHelicalLarge1]:NEEDS[Kerbalism,PlanetSideExplorationTechnologies]:AFTER[KerbalismDefault]
+{
+    !MODULE[ModuleGenerator] {}
+    MODULE
+    {
+        name = ProcessController
+        resource = _ECGen
+        title = Turbine
+        capacity = 20.0
+        running = false
+        toggle = true
+    }
+}
+@PART[Benjee10_MMSEV_baseTurbineHelicalLarge2]:NEEDS[Kerbalism,PlanetSideExplorationTechnologies]:AFTER[KerbalismDefault]
+{
+    !MODULE[ModuleGenerator] {}
+    MODULE
+    {
+        name = ProcessController
+        resource = _ECGen
+        title = Turbine
+        capacity = 20.0
+        running = false
+        toggle = true
+    }
+}
+@PART[Benjee10_MMSEV_baseTurbineHelicalTiny]:NEEDS[Kerbalism,PlanetSideExplorationTechnologies]:AFTER[KerbalismDefault]
+{
+    !MODULE[ModuleGenerator] {}
+    MODULE
+    {
+        name = ProcessController
+        resource = _ECGen
+        title = Turbine
+        capacity = 2.0
+        running = false
+        toggle = true
+    }
+}
+@PART[Benjee10_MMSEV_baseTurbineLowDensity]:NEEDS[Kerbalism,PlanetSideExplorationTechnologies]:AFTER[KerbalismDefault]
+{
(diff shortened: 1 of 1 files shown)


GameData/KerbalismConfig/Support/PlanetSideExplorationTechnologies.cfg
```diff
@@ -0,0 +1,101 @@
...
+    {
+        name = ProcessController
+        resource = _ECGen
+        title = Turbine
+        capacity = 15.0
+        running = false
+        toggle = true
+    }
+}
+@PART[Benjee10_MMSEV_baseTurbineMedDensity]:NEEDS[Kerbalism,PlanetSideExplorationTechnologies]:AFTER[KerbalismDefault]
```
Aebestach, line 101: It doesn't look right here

Aebestach:
Hi @Sandoleathy , thanks for putting this together. I checked the patch against the PlanetSideExplorationTechnologies part configs and the relevant Kerbalism modules. There are a few issues that should be sorted out before this is merged.

## Wind turbines

The Planetside turbines use `ModulePETTurbine`, not `ModuleGenerator`. There are no `ModuleGenerator` nodes on these parts, so this line does nothing:

```cfg
!MODULE[ModuleGenerator] {}
```

That leaves the original turbine running alongside the new `ProcessController`. Once the Kerbalism process is switched on, a loaded vessel can get power from both modules.

Replacing `ModulePETTurbine` outright would not be a good fix either. It handles deployment, wind speed, atmospheric efficiency, wind direction, damage, and repairs. A plain `_ECGen` process knows nothing about those conditions. It can therefore keep producing at a fixed rate while the turbine is retracted or broken, or even on an airless body.

This is also why copying `chargeRate` into `capacity` is only an approximation. The values in the PR match the turbines' nominal rates, but their actual output is modified by `atmEfficiencyCurve` and other conditions.

Kerbalism's `ProcessController` simply exposes a fixed amount of its pseudo-resource and uses its own `running` flag:

https://github.com/Kerbalism/Kerbalism/blob/e9a969353fe3a85a210d91b4565f5bcea5b06a40/src/Kerbalism/Modules/ProcessController.cs#L10-L18

https://github.com/Kerbalism/Kerbalism/blob/e9a969353fe3a85a210d91b4565f5bcea5b06a40/src/Kerbalism/Modules/ProcessController.cs#L122-L151

https://github.com/Kerbalism/Kerbalism/blob/e9a969353fe3a85a210d91b4565f5bcea5b06a40/src/Kerbalism/Modules/ProcessController.cs#L243 (`IsSituationValid() => true`)

It is not synchronized with `ModulePETTurbine`, so deploying a turbine does not start the Kerbalism process, and retracting or breaking it does not stop that process.

There are two coverage issues as well:

- `Benjee10_MMSEV_baseTurbineHelicalSmall` is missing. Its nominal output is 8 EC/s.
- The file ends at the selector for `Benjee10_MMSEV_baseTurbineMedDensity`, with no patch body. Its nominal output is 15 EC/s.

Planetside has eight turbine parts in total:

- Giant: 50 EC/s
- Helical Large A: 20 EC/s
- Helical Large B: 20 EC/s
- Helical Small: 8 EC/s
- Helical Tiny: 2 EC/s
- Low Density: 10 EC/s
- Medium Density: 15 EC/s
- High Density: 15 EC/s

Finally, `_ECGen` belongs to the default profile. These patches should require `ProfileDefault`; otherwise `ProcessController.Start()` may not find a matching process in the active profile.

Accurate background support would likely need dedicated integration for `ModulePETTurbine`, rather than a fixed `ProcessController`.

## Expandable habitat module 

The original deployment module is:

```cfg
MODULE
{
    name = ModuleAnimateGeneric
    CrewCapacity = 6
    animationName = atticDeploy
    ...
    restrictedNode = topNode
}
```

In Kerbalism, `Habitat.inflate` is the name of an animation, not a boolean. It should refer to `atticDeploy`; `inflate = true` looks for an animation literally named `true`.

`deployed` is not a `Habitat` field either. Kerbalism uses the persistent `state` field, which defaults to `enabled`. The intended initial setting would be `state = disabled`.

There is a second issue with crew capacity. The part has no top-level `CrewCapacity`; its six seats are declared inside `ModuleAnimateGeneric`. Kerbalism's generic habitat patch therefore does not match this part, and `Habitat` later reads a prefab crew capacity of zero:

https://github.com/Kerbalism/Kerbalism/blob/e9a969353fe3a85a210d91b4565f5bcea5b06a40/GameData/KerbalismConfig/System/Habitat.cfg#L1-L15

https://github.com/Kerbalism/Kerbalism/blob/e9a969353fe3a85a210d91b4565f5bcea5b06a40/src/Kerbalism/Modules/Habitat.cs#L244

https://github.com/Kerbalism/Kerbalism/blob/e9a969353fe3a85a210d91b4565f5bcea5b06a40/src/Kerbalism/Modules/Habitat.cs#L563-L569

As a result, the added `Habitat` does not properly own either the deployment or the six-seat capacity. The original animation module remains active as a separate controller, while the Habitat starts in its default enabled state.

A better starting point would be:

1. Set top-level `CrewCapacity = 6` before `KerbalismDefault`.
2. Let Kerbalism create the normal `Habitat`.
3. Patch it afterwards with `inflate = atticDeploy` and `state = disabled`.
4. Work out whether the original `ModuleAnimateGeneric` should be removed or synchronized.

The last point needs testing. Removing the stock module also removes its `restrictedNode = topNode` check, which Kerbalism's Habitat does not replace.

This patch should also require `FeatureHabitat`, so it does not add a Habitat when that feature is disabled.

## Command-module storage

The HDU command module does receive a `HardDrive` from Kerbalism's global science patch, so editing it with `@MODULE[HardDrive]` is valid when science is enabled.

Please add `FeatureScience` to the `:NEEDS` clause. The proposed values are technically valid, although 128 sample slots is much larger than the usual command-module capacity. A short balance comparison with a similar part would help justify that value.

## Minor point

`title = Turbine` should ideally use a localization key if the turbine process remains.

Aebestach:
Planetside Exploration Technologies likely requires code-level support rather than configuration alone, so I'll close this PR and open a new issue to address this later. Thank you for your contribution.

PR #991, opened 2026-01-21
Add suport for Feline utility rover

IvanSanchez:
This adds support for the drill and ISRU from the "Feline Utility Rovers" mod (https://spacedock.info/mod/1172/Feline%20Utility%20Rovers), by means of copy-pasting the config for the stock mini-drill and stock mini-isru.

I wish I could provide a cleaner configuration instead of copy-pasting a bunch of stuff, but my ModManager skills are next to nil.

The other rover parts from that mod seem to work just fine without any specific tweaks.

### DIFF
--- a/GameData/KerbalismConfig/Support/FelineUtilityRover.cfg
+++ b/GameData/KerbalismConfig/Support/FelineUtilityRover.cfg
@@ -1,0 +1,435 @@
+// Support for Lynx parts, from the Feline Utility Rover mod.
+//
+// Stats for minidrill and miniisru copy-pasted from stock.
+// Stats for habitats copy-pasted from Buffalo rover configs.
+
+
+// ============================================================================
+// Lynx Rear Mini-drill
+// ============================================================================
+@PART[Lynx_DrillRear]:NEEDS[FelineUtilityRover,ProfileDefault]:FOR[KerbalismDefault]
+{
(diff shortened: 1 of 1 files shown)


Aebestach:
@IvanSanchez 
Hi, I noticed a few issues, so I've moved it to draft for now—I'll mark it as "ready for review" once it's sorted out. It looks like the Kerbetrotter toggle gameplay from FUR was removed here, which might need some discussion: whether to follow Kerbalism's approach and drop some of FUR's native features, or implement code-level support for this toggling behavior. Also, since Kerbalism 3.40 added SystemHeat compatibility, I think FUR needs to be brought into the loop as well. Do you have time to continue maintaining these configs?

> My opinion: The switching capability comes with a high implementation cost in Kerbalism, so it's not really recommended. But EVA + engineer level can modify configurations, so I think it should be doable with just cfg.

IvanSanchez:
> Do you have time to continue maintaining these configs?

Alas, I have too many things on my plate (and I'm not a KSP modder, just a player), and I was unaware of FUR/Kerbetrotter/SystemHeat shenanigans. I'm afraid it'll be best to put this in the back burner for the time being.

Aebestach:
We already resolved this in #1074. Thanks again for submitting that PR—otherwise we wouldn't have known about the compatibility issue with FUR.

Issue #1013, opened 2026-04-15
[Feature request] Make repairing parts consume consume EVA repar kits


rocket-girl-e:
Repairable failures and servicing often feel kinda impactless on crewed mission. Just do an EVA and it's fine. Assuming sufficient nitrogen reserves, it's literally a non-problem. Making the repairs actually cost something on their own would make the failures feel more meaningful and real, and stock EVA repair kits are imo the perfect already existing thing for it. Feature like that would mean that you'd have to plan ahead, how much to take, is it worth repairing a system if it may mean that some failure later is not repairable and so on. 
Possibly add this as an optional thing in a difficulty menu so that people who don't want extra hurdle, or who update an already running game and haven't planned for this, can disable it.

R-T-B:
Added in latest release 3.32.

PR #1016, opened 2026-04-16
[Feature request] Make repairing parts consume EVA repair kits

KerballOne:
https://github.com/Kerbalism/Kerbalism/issues/1013


R-T-B:
This is something I have desired for a bit, thanks for taking the time to make it so.

PR #1018, opened 2026-04-26
Do not evaluate Reliability when that feature is disabled

siimav:
ROKerbalism configs do not use the Reliability feature and thus it would be nice to save many hundreds of milliseconds by not running the related code at all.


DRVeyl:
You like whitespace too much!  Totally could have written:
```
malfunction = Features.Reliability && Reliability.HasMalfunction(Vessel);
critical = Features.Reliability && Reliability.HasCriticalFailure(Vessel);
```

siimav:
And then you have 2 branches and two memory accesses which is totally suboptimal from the performance perspective!

PR #1030, opened 2026-05-21
Configure: dropdown UI with conflict highlights

chambm:
## Summary

Replaces the prev/next cycle arrows on each slot of the `Configure` PartModule with an expandable dropdown listing every unlocked setup. The cycle made it impossible to tell how many options were available, and rapid clicks could loop past the one you wanted; the dropdown shows the full list at once and a single click selects.

Screenshot 2026-05-21 173834

Screenshot 2026-05-21 161620

Particularly noticeable on RP-1 capsules / probes where the "Configure Experiments" Configure module can hold 4–8 slots out of a dozen unlocked experiments — the cycle was painful, the dropdown is much faster.

## What the user sees

Section title doubles as the expand/collapse toggle (▼ / ▲) when more setups are unlocked than slots. While expanded the selected setup's description is hidden — the option is in the list with its own row and the description would just be noise.

Padding entries (`None 2`, `None 3`, …) that KerbalismConfig generates so multi-slot modules can hold `None` in every slot are hidden from the dropdown; only a single `None` appears.

For experiment-style Configure modules the dropdown rows are colored so duplicates jump out:
- **Red (`#ff5555`)** — the option is already selected by another slot of *this* Configure module (picking it would run the same setup twice on this part).
- **Yellow (`#ffcc55`)** — the option's `Experiment` setup module has an `experiment_id` that's already running on a *different* part of the vessel.

The same colour is applied to the section title, so a problematic current selection stays visible when the dropdown is collapsed.

## Implementation notes

- `Configure.Render_panel` rewritten to call `AddSection` with a `click` callback that toggles per-slot expansion state stored in a new `HashSet<int> expanded_slots`, and to either render the setup's existing detail entries (collapsed) or the new option list (expanded). The detail-rendering code path is unchanged.
- Conflict colours come from `SetupHighlightColor(setup, slot)`. `IsSetupInOtherSlot` walks the local `selected` list; `CollectVesselExperimentIds` walks the editor's `EditorLogic.fetch.ship.parts` (or the live `vessel.parts`) and unions every other `Configure` module's selected setups' `Experiment`-type `id_value`s into a `HashSet<string>`. Computed once per `Window_body()` refresh and stashed in `vesselExperimentIds_cache` for the per-row lookup.
- `Panel.AddSection` gains an optional `click` callback (separate from the existing left/right arrow callbacks) that fires when the section title is clicked. Used to drive the dropdown toggle.
- `Panel.Render` makes the whole row of a clickable entry clickable (label *and* value, not just the value column) and draws a subtle translucent overlay over the row when the cursor is on it.
- `Styles.entry_container_wrap` (new) is a no-`fixedHeight` variant of `entry_container` used by clickable rows so long titles can wrap and grow the row vertically instead of squeezing into the gap between adjacent rows. A `GUILayout.MinHeight` floor matching the original `entry_container.fixedHeight` keeps single-line rows visually identical to before.
- `Tooltip.Render_tooltip` caps the tooltip width to the popup width so long descriptions wrap inside the window instead of clipping off the edges; when wrapping kicks in we widen to the full popup width so the wrap uses available horizontal space.

## Test plan

- [x] Editor: cycle arrows replaced by dropdown; click title to expand/collapse; click an option to select.
- [x] Editor: opening multiple slots and choosing the same setup colours the duplicate red on the other slot.
- [x] Editor: placing the same experiment-bearing setup on two parts colours the second yellow.
- [x] Editor: `None` and `None`-padding entries hidden except for the slot currently holding one.
- [x] Editor: collapsed section title keeps the conflict colour.
- [x] Long titles wrap inside the row without overlapping adjacent rows; non-wrapped rows look identical to before.
- [x] Long tooltips wrap inside the popup window.
- Flight (with `reconfigure = true`): same behaviour as editor. (Default Kerbalism behaviour is editor-only; flight access requires the part config to opt in via `reconfigure = <trait>` and a satisfying kerbal on the vessel.)

Drafted while I shake out any edge cases.

### DIFF
--- a/src/Kerbalism/Modules/Configure.cs
+++ b/src/Kerbalism/Modules/Configure.cs
@@ -26,6 +26,10 @@ namespace KERBALISM
 
 	public class Configure : PartModule, IPartCostModifier, IPartMassModifier, IModuleInfo, ISpecifics, IConfigurable
 	{
+		// dropdown colour markers for setups already present on the craft
+		const string COLOR_SAME_MODULE = "#ff5555";  // already selected by another slot of this Configure module
+		const string COLOR_SAME_VESSEL = "#ffcc55";  // experiment_id already running on a different part of the vessel
+
 		// config
 		[KSPField] public string title = string.Empty;           // short description (may be a #loc key; display via Localizer.Format)
 		[KSPField] public string configureId = string.Empty;     // stable MM-safe id (no #); use HAS[#configureId[...]] instead of HAS[#title[#loc]]
@@ -50,6 +54,8 @@ namespace KERBALISM
 		bool initialized;                                         // keep track of first configuration
 		CrewSpecs reconfigure_cs;                                 // in-flight reconfiguration crew specs
 		Dictionary<int, int> changes;                             // store 'deferred' changes to avoid problems with unity gui
+		HashSet<int> expanded_slots;                              // per-slot dropdown expansion state
+		HashSet<string> vesselExperimentIds_cache;                // populated each Window_body() refresh; consulted for yellow overlap highlight
 
 		// used to avoid infinite recursion when dealing with symmetry group
 		static bool avoid_inf_recursion;
@@ -101,6 +107,7 @@ namespace KERBALISM
 
 			// store configuration changes
 			changes = new Dictionary<int, int>();
+			expanded_slots = new HashSet<int>();
 		}
 
 
@@ -564,6 +571,11 @@ namespace KERBALISM
 				if (GetInstanceID() == 0) return;
 			}
 
+			// compute the vessel-wide set of running experiment_ids once per refresh; both
+			// the section titles and the expanded dropdown rows consult it for the yellow
+			// "already running elsewhere" highlight.
+			vesselExperimentIds_cache = CollectVesselExperimentIds();
+
 			// for each selected setup
 			for (int selected_i = 0; selected_i < selected.Count; ++selected_i)
 			{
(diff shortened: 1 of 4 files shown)


chambm:
Built `.kbin` attached to a preview release on my fork for anyone who'd like to try this without compiling:

https://github.com/chambm/Kerbalism/releases/tag/configure-dropdown-preview

**KSP 1.12 only** in this preview — single-version Debug build. Back up your existing `GameData/Kerbalism/Kerbalism112.kbin` and replace it with the one from the release.

Tested on KSP 1.12.5 with RP-1 + KerbalismConfig — Configure modules with `title = Experiments` (RP-1 capsules / probes) show the dropdown, with red highlights for setups already in another slot of the same module and yellow highlights for experiments already running on another part of the vessel.

Aebestach:
Report an issue: The goal of this PR is to add dropdown selection UI to the Configure module, but it modifies the shared Panel.cs / Styles.cs, so all windows using this set of Panel will be affected, including Body Info, etc., which may cause some UI elements to look unattractive. @chambm

chambm:
I actually realized that after making the PR, but I've been playing with it for weeks and still prefer it. Is there an example you've found where it changes something for the worse?

Actually the tooltips are still a problem. Sometimes they are truncated to the edges of the parent panel which is too restrictive. Not my earlier fix wasn't good, it made long tooltips take up the entire screen instead of wrapping to multiline.

Aebestach:
> I actually realized that after making the PR, but I've been playing with it for weeks and still prefer it. Is there an example you've found where it changes something for the worse?

But I didn't find any other problems

> Actually the tooltips are still a problem. Sometimes they are truncated to the edges of the parent panel which is too restrictive. Not my earlier fix wasn't good, it made long tooltips take up the entire screen instead of wrapping to multiline.

Yes, I’ve noticed this issue too—some are squeezed in and can’t be fully viewed. But perhaps there’s a better solution?

chambm:
Superceded by #1072

PR #1035, opened 2026-05-31
Fix invalid Emitter radiation value on NFE rtg-0625

Aebestach:
Move the ~0.031 rad/h note to a comment so PartLoader can parse the radiation field.


Aebestach:
However, rtg-0625 has been classified as legacy in the NFE and has been replaced by another reactor. Nevertheless, it is still important to note that a comment should be added here.

PR #1038, opened 2026-06-07
Fix missing Plants factor in planner

siimav:


R-T-B:
this seems so basic that I wonder how it could break anything.  Famous last words?

PR #1054, opened 2026-07-14
Allow supply and configure to add to existing resources

### LINKED ISSUE #1051: Existing resource on part is overwritten by Supply or Configure
When Kerbalism adds a resource to a part through either Supply or Configure, it overrides instead of adds to the matching resource if it already exists. This prevents different amount of supply resources to be added to parts on a part by part basis.

Using a += here: https://github.com/Kerbalism/Kerbalism/blob/master/src/Kerbalism/Lib.cs#L2698-L2703
does appear to fix the issue in the editor. RemoveResource function does use -=. Unsure the operational impact for parts in flight.

SerShrubert:
For https://github.com/Kerbalism/Kerbalism/issues/1051
Add optional parameter to Lib.AddResource that will allow existing resources to be added to, instead of overwritten. 

This option will be used in two cases:
1. When adding a resource defined in supply in SetupPod (Not on_rescue). This will be added to the existing resources on that part, so that life support resources can be configured on a per part basis. A 3 kerbal command pod could be configured to contain an extra 10 days of food for example. This adds some parity to the life support patches used by other mods for other life support systems.
2. Resources defined in the setup for a configure module. This does not apply to pseudoresources, only for the RESOURCE nodes inside a SETUP.

VAB testing:
Two 3 kerbal command pods with different amounts of food
Capture4

A configure option that adds additional amount of EC capacity. The extra capacity is correctly removed when switching to a different option

Capture1
Capture2
Capture3

I originally tested this using += in all cases, but this caused some issues with habitat and pseudoresources, mostly when undocking 2 command pods and kerbal EVA. The default behavior is retained for all other cases.

### DIFF
--- a/src/Kerbalism/Lib.cs
+++ b/src/Kerbalism/Lib.cs
@@ -2612,7 +2612,7 @@ namespace KERBALISM
 		/// <summary> Adds the specified resource amount and capacity to a part,
 		/// the resource is created if it doesn't already exist </summary>
 		///<summary>poached from https://github.com/blowfishpro/B9PartSwitch/blob/master/B9PartSwitch/Extensions/PartExtensions.cs
-		public static PartResource AddResource(Part p, string res_name, double amount, double capacity, bool append_amount = false)
+		public static PartResource AddResource(Part p, string res_name, double amount, double capacity, bool addToExisting = false)
 		{
 			var reslib = PartResourceLibrary.Instance.resourceDefinitions;
 			// if the resource is not known, log a warning and do nothing
@@ -2652,25 +2652,21 @@ namespace KERBALISM
 			}
 			else
 			{
-				if (append_amount) // Add to existing resource instead of overriding it
+				PartResource simulationResource = p.SimulationResources?[resourceDefinition.name];
+
+				if (addToExisting)
 				{
 					resource.maxAmount += capacity;
-
-					PartResource simulationResource = p.SimulationResources?[resourceDefinition.name];
-					if (simulationResource != null) simulationResource.maxAmount += capacity;
-
-					resource.amount += amount;
+					resource.amount = Math.Min(resource.amount + amount, resource.maxAmount);
 				}
-				else // Default behavior
+				else
 				{
 					resource.maxAmount = capacity;
-
-					PartResource simulationResource = p.SimulationResources?[resourceDefinition.name];
-					if (simulationResource != null) simulationResource.maxAmount = capacity;
-
 					resource.amount = amount;
 				}
 
+				if (simulationResource != null)
+					simulationResource.maxAmount = resource.maxAmount;
 			}
 
 			return resource;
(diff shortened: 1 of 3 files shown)


SerShrubert:
This change was created in support of my BDB compatibility patch https://github.com/Kerbalism/Kerbalism/issues/1069

Aebestach:
Confirmed in VAB with a stock Mk1-3 that already has a base Nitrogen tank:

1. Enable Pressure Control → Nitrogen becomes 100 + 993.9
2. Lower Nitrogen to something like 547 / 1093.9
3. Pressure Control -> None/Scrubber/or any other non Pressure Control

Result: 0 / 100
Expected: 100 / 100 (only remove the configure capacity, keep remaining contents up to the base tank)

Append looks good for capacity. When disabling the setup, though, RemoveResource still subtracts the configure amount, so any shared base resource can get wiped. It might be safer to remove capacity only, clamp amount to >= 0, and keep SimulationResources in sync.


```cfg

// A simple, quick test script
@PART[mk1-3pod]:NEEDS[ProfileDefault]:AFTER[KerbalismDefault]
{
	RESOURCE
	{
		name = Nitrogen
		amount = 100
		maxAmount = 100
	}
}
```

SerShrubert:
> Append looks good for capacity. When disabling the setup, though, RemoveResource still subtracts the configure amount, so any shared base resource can get wiped. It might be safer to remove capacity only, clamp amount to >= 0, and keep SimulationResources in sync.

Should we remove line 2675 resource.amount -= amount; and then let amount be clamped to capacity as usual?

Aebestach:
> > Append looks good for capacity. When disabling the setup, though, RemoveResource still subtracts the configure amount, so any shared base resource can get wiped. It might be safer to remove capacity only, clamp amount to >= 0, and keep SimulationResources in sync.
> 
> Should we remove line 2675 resource.amount -= amount; and then let amount be clamped to capacity as usual?

https://github.com/Kerbalism/Kerbalism/compare/master...Aebestach:Kerbalism:fix/pr-1054-resource-append
This should help you out

SerShrubert:
> > > Append looks good for capacity. When disabling the setup, though, RemoveResource still subtracts the configure amount, so any shared base resource can get wiped. It might be safer to remove capacity only, clamp amount to >= 0, and keep SimulationResources in sync.
> > 
> > 
> > Should we remove line 2675 resource.amount -= amount; and then let amount be clamped to capacity as usual?
> 
> [master...Aebestach:Kerbalism:fix/pr-1054-resource-append](https://github.com/Kerbalism/Kerbalism/compare/master...Aebestach:Kerbalism:fix/pr-1054-resource-append) This should help you out

Cherry picked these changes with $ git cherry-pick --strategy=recursive -X theirs 66c80f30a021bd391e088ac3d244e3e3bc4ad0fe

Aebestach (approved):

R-T-B (approved):
Seems reasonable.  I'll merge it.  Good luck with the compatibility bridge.

Issue #1031, opened 2026-05-22
NaN in experiments


ScriBanana:
Hello,

I regularly get NaNs on experiments, with a time to completion of perpetual. The logs ([KSP.log](https://github.com/user-attachments/files/28090571/KSP.log)) give a bunch of `Experiment.RunningUpdate warpDrive.FileCapacityAvailable() returned NaN`. Doesn't appear to affect sample experiments, no relation to available data storage on the ship nor connection to DSN.

- The issue persists across saves, including with new saves.
- It affects all experiments, across all ships.
- It sometimes gets randomly fixed by rebooting the game, which makes troubleshooting very unpredictable.
- I use a fair amount of mods (see below).
- I hadn't had the issue until I updated Kerbalism to 3.32, updated a few other mods and installed some more (see below as well).
- I tried reverting Kerbalism to 3.31 and other mods to the previous versions, but it did not fix the issue (I tried different combinations, such as reverting only Kerbalism, only Probes Before Crew, Kopernicus, etc. to no avail.) Again, difficult to properly troubleshoot as restarting the game sometimes solves the issue.
- There is this line prior to the bunch of NaNs: `[Kerbalism] Sim.get_DataRateDampingExponent Calculated DataRateDampingExponent: NaN (max. DSN range: 25000000373, strength at 2 AU: 0.000)`. I saw #721, but the issue persists across saves, regardless of DSN strength. Don't know if it isn't a red herring.

I know this is not extremely precise, but I can experiment some more or provide with more details on request.

<details>
<summary>Full mod list</summary>

- 4kSP_Expanded (4kSPExpanded 0.2.2)
- Antenna Helper Next (AntennaHelperNext 1.0.8)
- Astrogator (Astrogator v1.0.2)
- Avalanche Plumes (AvalanchePlumes v1.0.5.1)
- B9 Part Switch (B9PartSwitch v2.21.0.4)
- BetterTimeWarpContinued (BetterTimeWarpCont 2.3.14.1)
- Burst PQS (BurstPQS 0.1.22)
- Chatterer (Chatterer 0.9.99)
- Chatterer Extended (ChattererExtended 0.6.2)
- ClickThrough Blocker (ClickThroughBlocker 1:2.1.10.22)
- Community Resource Pack (CommunityResourcePack v112.0.1)
- Community Tech Tree (CommunityTechTree 1:3.4.5)
- Contract Configurator (ContractConfigurator v2.13.0.0)
[...]

Issue #721, opened 2021-03-07
Consistent NaN in Science Experiments -- Kerbalism 3.11 & 3.12 / KSP 1.10.1.2939


romyers:
Appears to be pretty much exactly what's described in #592 . Science experiments seem to behave properly on the launch pad, but the stored data goes NaN immediately on takeoff. 

This may be associated with the following line in my log file, which repeats continuously (I think after takeoff, if Mechjeb's logging is to be believed):

`[ERR 11:26:06.806] [Kerbalism] Experiment.RunningUpdate warpDrive.FileCapacityAvailable() returned NaN`

I'm also seeing the following NaNs for what appears to be an asteroid's orbital parameters:

```
[LOG 11:21:35.092] ObT : NaN
M : NaN
E : NaN
V : NaN
Radius: NaN
vel: [NaN, NaN, NaN]
AN: [NaN, NaN, NaN]
period: Infinity

[WRN 11:21:35.092] [OrbitDriver Warning!]: Ast. YRJ-552 had a NaN Orbit and was removed.
```

There are no other NaNs in the log file.

This has happened consistently on my rather heavily modded game. I have not yet attempted to reproduce the bug without my full modset. Within a single session, there appears to be some vessel-dependence, even across multiple uses of the same vessel with Kerbal Construction Time (one vessel never experienced the NaN bug, despite other vessels exhibiting it in the same session, but unfortunately I discarded the vessel after my next reload).

Attached are my log file and module manager configCache, and a video of the flight the log files are associated with. This flight is the first flight on a fresh save file, and the log spans the creation of that save file:

Log/configCache: [Kerbalism NaN.zip](https://github.com/Kerbalism/Kerbalism/files/6097551/Kerbalism.NaN.zip)

Video: [https://drive.google.com/file/d/1hgm3rwwcFVQ1YKmSBJrbQ2L1D45Nyk4R/view?usp=sharing](url)

Any light you could shed on this issue would be appreciated! If necessary, I can try uninstalling my mods and gradually adding them back in, but this would take me more time than I have right now so it would have to wait.

Thanks!!!

romyers:
Possibly associated with Kopernicus or Beyond Home -- as a cursory test, I removed those mods, and the bug did not appear on the next attempt. But I only tried one launch; hardly exhaustive.

gotmachine:
Thanks for the report.
[...]

PR #1056, opened 2026-07-15
Fix NaN damping exponent breaking science experiments (CommNet)

Aebestach:
#1031 When DSN can't reach 2 AU, `DataRateDampingExponent` auto-calc hits `Math.Log(..., 0)` and returns NaN. That poisons comm rates and experiment storage capacity for the rest of the session.

This adds a fallback (6.0 for CommNet, 2.4 for RT) and skips invalid cached values. Same failure mode as #721 / #914, but CommNet path was missing the guard.

_fallback is a safe value to prevent NaN, users of distant galaxies should still manually modify `DampingExponentOverride`_

### DIFF
--- a/src/Kerbalism/Sim.cs
+++ b/src/Kerbalism/Sim.cs
@@ -1468,12 +1468,17 @@ namespace KERBALISM
 		#endregion
 
 		#region SIGNAL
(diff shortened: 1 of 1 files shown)


R-T-B:
This looks like a solid workaround for the core logical issue, good work.

PR #1059, opened 2026-07-15
Clarify MannedOrbit contract text to match vessel-based check

### LINKED ISSUE #1028: MannedOrbit contract text a bit misleading
This is just a little thing, but at least in the English the contract description for the MannedOrbit contract seems to imply that completing the contract requires keeping a *Kerbal* in orbit for 30 days. This is not the case, as the code does the 30-day check based on the time the crewed *vessel* has been above the atmosphere. This means edge cases such as mine (see below) do not complete as expected.

I first tried to complete this contract by sending Valentina into orbit around Kerbin with sufficient supplies for 30 days. It was only after being up there for a few days that I realized her stress levels were not going to let her stay up there that long, so I launched a new larger craft with new supplies, Bob, and some empty seats three days later; once in orbit I rendezvoused with Valentina's craft and transferred her into the new craft so they weren't alone and stress accumulation would be lower. I figured that once Valentina had been in orbit for 30 days between her time on the two vessels the contract would complete, but it has not and I need to wait about 2.5 days more.

I don't think how the contract works needs to be changed necessarily, just that it could be worded better about what one has to do: keep a crewed vessel in orbit for 30 days.

Aebestach:
Improvements to the text for #1028


Aebestach:
Once #1062 is merged, I'll make some minor tweaks to this

PR #1072, opened 2026-07-17
Add Configure dropdown UI with opt-in panel and tooltip fixes (Improvements to #1030)

Aebestach:
Based on my understanding of #1030, I made improvements on top of @chambm work by applying the scrollbar only to the Configure section rather than the global UI.

### DIFF
--- a/GameData/Kerbalism/Localization/de.cfg
+++ b/GameData/Kerbalism/Localization/de.cfg
@@ -1078,6 +1078,7 @@ Localization
     #KERBALISM_Experiment_issue = Problem : // "issue :"
     #KERBALISM_Experiment_sciencevalue = Wissenschaftswert : // "science value :"
     #KERBALISM_Experiment_completion = Fertigstellung : // "completion :"
+    #KERBALISM_Process_capacity = Prozesskapazität : // "Process capacity :"
     #KERBALISM_SolarPanel_deployable = Solarmodul (entfaltbar) // "solar panel (deployable)"
     #KERBALISM_SolarPanel_nonretractable = Solarmodul (nicht entfaltbar) // "solar panel (non retractable)"
     //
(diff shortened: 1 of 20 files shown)


Aebestach:
[Kerbalism 3.41-beta.2.zip](https://github.com/user-attachments/files/30118864/Kerbalism.3.41-beta.2.zip)
@chambm Hi there, I have a test version here—would you mind giving it a try?

chambm:
Yes, this is better. Would it be better to move the tooltip to the mouse? Then you have to worry about it going off the screen though and handle that.

Aebestach:
> Yes, this is better. Would it be better to move the tooltip to the mouse? Then you have to worry about it going off the screen though and handle that.

I'll look into it.

Aebestach:
@chambm 
The current tooltip strategy works as follows:
1. By default, position it to the lower right of the cursor with 12–20px of spacing, so it doesn't obscure the current option.
2. If there's insufficient space on the right, flip it to the left.
3. If there's insufficient space below, flip it to the top.
4. Finally, clamp against screen edges as a last resort.

---

Also fixed two hard-coded fields.

---
Also fixed a tooltip that had inexplicable line breaks for some reason.
https://github.com/Kerbalism/Kerbalism/blob/618e09ccd720db79ab39b0678c1bc42fe8e535ab/src/Kerbalism/UI/ConnManager.cs#L41

R-T-B (approved):
looks good to me.

SerShrubert:
With the current implementation in #1072 the bottom item is partially cut off, and the scroll bar visible even when there are not many modules. The size of the window is also not fixed, it will expand to fit all but one option. The scroll bar should be disabled and all entries visible when below a certain amount of amount of entries. 
Attached some examples with various amounts of options.
Configure
Kerbalism2
Kerbalism3
Kerbalism4

Aebestach:
> With the current implementation in #1072 the bottom item is partially cut off, and the scroll bar visible even when there are not many modules. The size of the window is also not fixed, it will expand to fit all but one option. The scroll bar should be disabled and all entries visible when below a certain amount of amount of entries. Attached some examples with various amounts of options. Configure Kerbalism2 Kerbalism3 Kerbalism4

Consider showing up to five items without a scrollbar; enable scrolling for five or more. I'll be making some adjustments to this section over the next few days.

PR #1075, opened 2026-07-18
Localize Auto device display names and experiment icon tooltips

Aebestach:
Keep English Name values for stable device Id hashing across languages.


Aebestach:
I always seem to notice unlocalized content when I least expect it

PR #1117, opened 2026-07-23
Probes/DMOS science support and SystemHeat radiators

### LINKED ISSUE #465: ProbesPlus support
The stock experiment parts in probesplus work, but the dmagic parts have the instant science behavior from before 3.0

### LINKED ISSUE #885: DMOS asteroid science not transmitting
DMOS (DMagic Orbital Science) experiments, at least Laser Scan, doesn't work when attached to an asteroid.

The experiment result window shows up correctly, but the data is not collected by Kerbalism (i.e. Data tab remains empty and nothing is transmitted).

Here is what the log shows:
`[Kerbalism] ScienceDB.GetSubjectDataFromStockId Could not parse the SubjectData from subjectId 'dmlaserblastscan@AsteroidSrfLandedCarbonaceous14393197' : the body 'Asteroid' doesn't exist`
image

Aebestach:
1. Fix DMOS asteroid subjects not storing/transmitting (#885)
2. Convert Probes / remapped DMOS probe experiments to Kerbalism Experiment (#465)
3. Add Probes science support and SystemHeat radiator patches (Louvers, ESM, Vorona wings)

### DIFF
--- a/GameData/KerbalismConfig/Support/CoatlAerospace_Science.cfg
+++ b/GameData/KerbalismConfig/Support/CoatlAerospace_Science.cfg
@@ -1,0 +1,649 @@
+// ============================================================================
+// Coatl Aerospace (ProbesPlus / GroundOps) - Kerbalism science support
+//
+// Convert Coatl ModuleScienceExperiment / DMModuleScienceAnimateGeneric instruments
+// to Kerbalism Experiment. Science points follow Coatl ScienceDefs baseValue.
+// Data size = stock baseValue * dataScale (Mb). EC is modest (original parts had none).
+//
+// Multi-experiment parts (mer_rsp, ca_sciBoom, ca_vor_camera, ca_stereoBoom, …) keep
+// one Experiment module per experimentID — simultaneous slots on the same part.
+//
+// DMOS remaps (magScan / rpwsScan / scopeScan on DM* modules) stay handled by
+// DMagicOrbitalScience_Science.cfg. This file covers ca_* IDs and MSE+scopeScan
+// leftovers after Coatl's DMagicOrbitalScience.cfg remaps.
+// ============================================================================
+
+@KERBALISM_EXPERIMENT_VALUES:NEEDS[FeatureScience]
+{
+	CoatlAerospace
+	{
+		// ScienceDefs: baseValue * dataScale → size; value = baseValue
+		ca_gammaRay
+		{
+			ECCost = 0.05
+			size = 6
+			value = 6
+			duration = 600								// 10 min
+			requirements =
+			ResourceRates =
+		}
+		ca_rpws
+		{
+			ECCost = 0.08
+			size = 14
+			value = 7
+			duration = 1800								// 30 min
+			requirements =
+			ResourceRates =
+		}
+		ca_orbitalScope
+		{
+			ECCost = 0.05
+			size = 18
+			value = 6
+			duration = 900								// 15 min
+			requirements = SunAngleMax:55
+			ResourceRates =
+		}
+		ca_mag
+		{
+			ECCost = 0.05
+			size = 5
+			value = 5
+			duration = 900								// 15 min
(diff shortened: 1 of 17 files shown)


GameData/KerbalismConfig/Support/CoatlAerospace_Science.cfg
```diff
@@ -0,0 +1,649 @@
...
+	MODULE
+	{
+		name = Experiment
+		experiment_id = ca_SiteSurvey
+		allow_shrouded = False
+	}
+	!MODULE:HAS[#experimentID[ca_SiteSurvey],~name[Experiment]] {}
+}
+
+@PART[*]:HAS[@MODULE:HAS[#experimentID[ca_soilScoop]]]:NEEDS[FeatureScience]
```
SerShrubert, line 333: The soil scoop on Surveyor/Landvermesser isn't intended to collect a sample. It only moves the soil around to be photographed.

GameData/KerbalismConfig/Support/CoatlAerospace_Science.cfg
```diff
@@ -0,0 +1,649 @@
...
+		ca_orbitalScope
+		{
+			ECCost = 0.05
+			size = 18
+			value = 6
+			duration = 900								// 15 min
+			requirements = SunAngleMax:55
+			ResourceRates =
+		}
+		ca_mag
```
SerShrubert: This should be removed in favor of the stock magnetometer experiment.

SerShrubert:
Most of the ProbesPlus experiments get converted to BDB experiments by ProbesPlus's own compatibility patch.
The patches to the parts should make sure to run after the patches in Bluedog_DB.cfg, so they don't get applied out of order.
Would it make sense to use the same values for the experiments that are identical between ProbesPlus and BDB? The durations I used were on the longer side, but they are still up for review.

Aebestach:
> Most of the ProbesPlus experiments get converted to BDB experiments by ProbesPlus's own compatibility patch.
> The patches to the parts should make sure to run after the patches in Bluedog_DB.cfg, so they don't get applied out of order.
> Would it make sense to use the same values for the experiments that are identical between ProbesPlus and BDB? The durations I used were on the longer side, but they are still up for review.

I haven't used ProbesPlus before. Once I've addressed your feedback, I'll merge it in. After that, you can make any further changes through the BDB PR—this PR here is solely for completing the Kerbalism compatibility work. We also need to keep in mind players who don't use BDB.

SerShrubert:
> > Most of the ProbesPlus experiments get converted to BDB experiments by ProbesPlus's own compatibility patch.
> > The patches to the parts should make sure to run after the patches in Bluedog_DB.cfg, so they don't get applied out of order.
> > Would it make sense to use the same values for the experiments that are identical between ProbesPlus and BDB? The durations I used were on the longer side, but they are still up for review.
> 
> I haven't used ProbesPlus before. Once I've addressed your feedback, I'll merge it in. After that, you can make any further changes through the BDB PR—this PR here is solely for completing the Kerbalism compatibility work. We also need to keep in mind players who don't use BDB.

Most of these experiments are a fallback for when BDB isn't installed. I can update their values in the BDB PR if they are determined to be reasonable.
I will verify the compatibility on my end after this PR is merged and address them in the BDB PR if necessary.

Aebestach:
@SerShrubert I made some revisions—could you take a look and let me know if this works?

SerShrubert:
> @SerShrubert I made some revisions—could you take a look and let me know if this works?

Looks good at first glance, but I won't be able to test personally until tonight or tomorrow.
With the animations, I left the scienceAnimateGeneric modules in until I use a universal patch to copy over the animation. I think this will avoid needing to patch parts that have or don't have animations separately. Alternatively can HAS be applied to the left hand side of an assignment operator if the right hand side doesn't match any variables? This is the error I ran into when originally getting animations to work.

Issue #1089, opened 2026-07-19
Space situation returns too much science in Solar orbit


SerShrubert:
Issue: The "Space" science situation currently uses the InSpaceLow science modifier. The default modifier for InSpaceLow around the Sun in the stock system is 11x, which is quite high as this situation is expensive to get to. The Space situation is reasonable until you get to solar orbit, then suddenly have a large amount of science coming in. Its unknown if this is intentional or not. Solar experiments restricted to Suns that are expecting a large science modifier should have their baseValue increased instead. 
geigerCounter, dmSIGINT, dmReconScan and dmSolarParticles use this situation. dmSIGINT and dmReconScan don't make sense for solar orbit in the first place. dmSolarCollector already has a high value so doesn't need the InSpaceLow boost, and would probably benefit from using InSpaceLow and InSpaceHigh instead to reward capturing samples from InSpaceLow anyway. geigerCounter has such a low baseValue that it benefits from getting extra science once in solar orbit. However it already has results for both SunInSpaceLow and SunInSpaceHigh so this doesn't seem intentional.

Option 1: Use the InSpaceHigh modifier instead. This will have slightly lower science value for all bodies other than the homeworld.
Option 2: Use InSpaceHigh modifier for suns, InSpaceLow for everything else.
Option 3: Do nothing, be mindful of experiments that use Space modifier.

Aebestach:
Handled in #1118

PR #1118, opened 2026-07-23
Use the SpaceHigh modifier for Space/Global for stars (#1089)

SerShrubert:
For https://github.com/Kerbalism/Kerbalism/issues/1089
The InSpaceLow modifier for Sun in stock solar system is 11x. Since this is a difficult situation to get to it is not reasonable to use it for the Space and Global situations.
Retain existing modifier for all other bodies.
This affects Kerbalism's GeigerCounter experiment. This is a reduction of 36 science for each of the virtual biomes that apply to suns (NoBiome, Storm, Interstellar)


Aebestach:
Looks like dmSolarParticles is affected too, but this PR makes sense.

Aebestach (approved):

PR #1130, opened 2026-07-26
Align Kerbalism calendar units with KSP dateTimeFormatter

### LINKED ISSUE #873: [Meta-Isssue] Custom day/year duration isn't well handled with some planet packs
Kerbalism has an automatic detection of the calendar parameters (hours per day and days per year) based on analysing the home body rotation period and orbit. The resulting parameters can end up being weird, incorrect or even buggy in some corner cases (tidally locked or short rotation period homeworlds). The parameters are also used inconsistently in some cases to have dynamically adjusted durations.

Ideally, this should likely be refactored to use Kronometer (or just parsing the Kronometer configs), which is the defacto standard used by planet packs to handle that specific issue.

Related issues :
#869
#822

Aebestach:
## Description

_**Try to resolve issue #873 and its related problems; testing is needed.**_

1. Split calendar handling into **display** and **gameplay**:
   - **Display / UI** (`HumanReadableDuration`, `HumanReadableRate`, timestamps): follow `KSPUtil.dateTimeFormatter` (stock Kerbin Time setting or Kronometer).
   - **Gameplay** (contracts, reliability, breakdown, solar cycle, RTG, etc.): in stock, use the home body’s physical day/year and **ignore** `GameSettings.KERBIN_TIME`; if a custom calendar mod replaces the formatter (e.g. Kronometer), follow that calendar.
2. Stop caching calendar lengths, since the formatter can be replaced during main-menu startup.
3. Add `MinuteSeconds` / `HourSeconds` / `DaySeconds` / `YearSeconds` helpers and use them for duration/rate formatting and year/day conversions.

### Known limitation

`IDateTimeFormatter` exposes `Minute` / `Hour` / `Day` / `Year` as **integers** and does not expose a custom “second” length. Kerbalism therefore assumes 1 display second = 1 game second. Fractional unit lengths from mods like Kronometer are truncated by that API; supporting non-standard second lengths would require going outside the public formatter interface.

### DIFF
--- a/src/Kerbalism/Background.cs
+++ b/src/Kerbalism/Background.cs
@@ -521,7 +521,7 @@ namespace KERBALISM
 
 			double power = Lib.ReflectionValue<float>(radioisotope_generator, "BasePower");
 			double half_life = Lib.ReflectionValue<float>(radioisotope_generator, "HalfLife");
-			double mission_time = v.missionTime / (3600.0 * Lib.HoursInDay * Lib.DaysInYear);
+			double mission_time = v.missionTime / Lib.YearSeconds;
 			double remaining = Math.Pow(2.0, (-mission_time) / half_life);
 			ec.Produce(power * remaining * elapsed_s, ResourceBroker.RTG);
 		}
--- a/src/Kerbalism/Contracts/MannedOrbit.cs
+++ b/src/Kerbalism/Contracts/MannedOrbit.cs
@@ -85,7 +85,7 @@ namespace KERBALISM.CONTRACTS
 				if (!vd.IsSimulated) continue;
 				bool manned = vd.CrewCount > 0;
 				bool in_orbit = Sim.Apoapsis(v) > v.mainBody.atmosphereDepth && Sim.Periapsis(v) > v.mainBody.atmosphereDepth;
-				bool for_30days = v.missionTime > 60.0 * 60.0 * Lib.HoursInDay * 30.0;
+				bool for_30days = v.missionTime > Lib.DaySeconds * 30.0;
 				if (manned && in_orbit && for_30days)
 				{
 					SetComplete();
--- a/src/Kerbalism/Lib.cs
+++ b/src/Kerbalism/Lib.cs
@@ -359,7 +359,9 @@ namespace KERBALISM
 
 		///<summary>
 		/// Display / UI calendar lengths in seconds.
-		/// Follows KSPUtil.dateTimeFormatter (stock Kerbin Time setting or Kronometer).
+		/// Stock: GameSettings.KERBIN_TIME (6h×426d vs 24h×365d) — stock dateTimeFormatter.Day/Year
+		/// are home-body physical periods and do not flip with that setting.
+		/// Custom calendar mods (Kronometer, etc.): follow their dateTimeFormatter.
 		/// Don't cache: the formatter can be replaced during main-menu startup.
 		///</summary>
 		private static void DisplayCalendarLengths(out double minute, out double hour, out double day, out double year)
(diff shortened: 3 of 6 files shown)


SerShrubert:
Solar cycle, breakdown probability, reliability and the manned orbit contract also use Lib.HoursInDay and Lib.DaysInYear. In the previous implementation the user setting for 24hr clock could affect these, but I do not believe that is desirable. With Kronometer installed I believe this setting does not do anything, but with a stock system this setting will have gameplay impact, when it should be purely visible.

Should GameSettings.KERBIN_TIME be completely ignored? Human readable duration will be on Kerbin time but unexpected gameplay changes won't happen. Is KERBIN_TIME only used as a fallback for the new initialization fallback?

Edit for clarification: This has gameplay impact. How does the Kerbin Time setting affect the stock system without Kronometer, and are there any planet packs with weird Kronometer settings that may have unexpected results? Tidally locked planets with really long days? Homeworlds in the outer solar system?

MorePortal:
I should note that, while Kronometer works with doubles internally (`TimeUnits` class in [KronometerLoader.cs](https://github.com/linuxgurugamer/Kronometer/blob/master/src/KronometerLoader.cs#L103)), 
`KSPUtil.dateTimeFormatter` returns integers when asked for time units (at the very end of [ClockFormatter.cs](https://github.com/linuxgurugamer/Kronometer/blob/master/src/ClockFormatter.cs#L812))

(usually calendars have an integer number of 'seconds' for each time unit, but the 'second' itself can be any value - making every other unit fractional. And even if it uses usual second, minutes and hours, the day length and the year length may be fractional)

Aebestach:
> I should note that, while Kronometer works with doubles internally (`TimeUnits` class in [KronometerLoader.cs](https://github.com/linuxgurugamer/Kronometer/blob/master/src/KronometerLoader.cs#L103)), `KSPUtil.dateTimeFormatter` returns integers when asked for time units (at the very end of [ClockFormatter.cs](https://github.com/linuxgurugamer/Kronometer/blob/master/src/ClockFormatter.cs#L812))
> 
> (usually calendars have an integer number of 'seconds' for each time unit, but the 'second' itself can be any value - making every other unit fractional. And even if it uses usual second, minutes and hours, the day length and the year length may be fractional)

@MorePortal 

Good catch. Yes — IDateTimeFormatter exposes Minute/Hour/Day/Year as int, so any fractional lengths from Kronometer get truncated at that API boundary; Kerbalism can’t recover the original doubles from there.

In practice the loss is usually < 1s per unit. If truncation yields 0/invalid values, we already reject them and fall back to the stock/home-body calendar. I don’t think we need a special workaround beyond that unless a pack hits a real edge case.

Aebestach:
> Solar cycle, breakdown probability, reliability and the manned orbit contract also use Lib.HoursInDay and Lib.DaysInYear. In the previous implementation the user setting for 24hr clock could affect these, but I do not believe that is desirable. With Kronometer installed I believe this setting does not do anything, but with a stock system this setting will have gameplay impact, when it should be purely visible.
> 
> Should GameSettings.KERBIN_TIME be completely ignored? Human readable duration will be on Kerbin time but unexpected gameplay changes won't happen. Is KERBIN_TIME only used as a fallback for the new initialization fallback?
> 
> Edit for clarification: This has gameplay impact. How does the Kerbin Time setting affect the stock system without Kronometer, and are there any planet packs with weird Kronometer settings that may have unexpected results? Tidally locked planets with really long days? Homeworlds in the outer solar system?

@SerShrubert 

Yeah, fair point — that one’s on us.

`KERBIN_TIME` isn’t just a fallback right now. In stock, once the formatter is up, solar cycle / breakdown / reliability / manned orbit all follow it. Flip Kerbin Time off and you’re on Earth calendar lengths, so gameplay shifts with a display toggle. That’s not what we want.

Kronometer is different: then we *should* follow the custom calendar. That’s the whole point of #873.

Plan:
- UI / human-readable time → keep using the formatter
- Gameplay day/year → don’t track the stock Kerbin Time switch; prefer the home world’s real day/year. If Kronometer (or similar) replaces the calendar, follow that

Weird cases (tidal lock, outer-system homeworlds): fine if the pack sets that via Kronometer; otherwise stick to physical home-body periods.

I’ll update the PR accordingly.

Aebestach:
[Kerbalism 3.41 Pre-release.zip](https://github.com/user-attachments/files/30651180/Kerbalism.3.41.Pre-release.zip)
@SerShrubert Could you help me test this?

SerShrubert:
> [Kerbalism 3.41 Pre-release.zip](https://github.com/user-attachments/files/30651180/Kerbalism.3.41.Pre-release.zip) @SerShrubert Could you help me test this?

On stock system, RTG half-life appears as 8yr 145d with both Kerbin time and 24hr time. If the RTG decay constant is the same then one of the human readable displays is wrong. Reliability mtbf display time appears to work correctly. Human readable duration for supply appears to work correctly.
Tested with JNSQ and Kronometer and all human readable durations appear correct.
I don't have a good way to test solar cycle duration or stress breakdowns, which are affected by the calendar units.

Aebestach:
The new push should have fixed it

PR #1147, opened 2026-08-02
Add support for biome values to science situation

SerShrubert:
Add support for per-biome science multipliers allowed by Kopernicus. Many planet packs modify the biome science multiplier.

Testing:
Gave Oceans a multiplier of 0.5, Hot Deserts 0.25.
Kerbalism

### DIFF
--- a/src/Kerbalism/Science/ScienceSituation.cs
+++ b/src/Kerbalism/Science/ScienceSituation.cs
@@ -113,7 +113,7 @@ namespace KERBALISM
 			return true;
 		}
 
-		public static float BodyMultiplier(this ScienceSituation situation, CelestialBody body)
+		public static float BodyMultiplier(this ScienceSituation situation, CelestialBody body, CBAttributeMapSO.MapAttribute biome = null)
 		{
 			float result = 0f;
 			switch (situation)
@@ -141,6 +141,14 @@ namespace KERBALISM
 			{
 				Lib.Log("Science: invalid/unknown situation " + situation.ToString(), Lib.LogLevel.Error);
 				return 1f; // returning 0 will result in NaN values
+			}
+
+			if (biome != null)
+			{
+				if (biome.value > 0f)
+				{
+					result *= biome.value;
+				}
 			}
 			return result;
 		}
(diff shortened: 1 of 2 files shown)


src/Kerbalism/Science/ScienceSituation.cs
```diff
@@ -113,7 +113,7 @@ public static bool IsAvailableOnBody(this VirtualBiome virtualBiome, CelestialBo
 			return true;
 		}
 
-		public static float BodyMultiplier(this ScienceSituation situation, CelestialBody body)
+		public static float BodyMultiplier(this ScienceSituation situation, CelestialBody body, CBAttributeMapSO.MapAttribute biome)
```
Aebestach: `BodyMultiplier` is public. It's recommended to write it as `CBAttributeMapSO.MapAttribute biome = null`, which preserves the two-parameter call and keeps it compilable.

src/Kerbalism/Science/ScienceSituation.cs
```diff
@@ -142,6 +142,14 @@ public static float BodyMultiplier(this ScienceSituation situation, CelestialBod
 				Lib.Log("Science: invalid/unknown situation " + situation.ToString(), Lib.LogLevel.Error);
 				return 1f; // returning 0 will result in NaN values
 			}
+
+			if (biome != null)
+			{
+				if (biome.value > 0f)
```
Aebestach, line 148: Is there ever a case where `biome.value = 0`? My suggestion: when it equals 0, always multiply first; if the result is 0, then fall back to the existing invalid situation handling.

Aebestach (changes requested):

src/Kerbalism/Science/ScienceSituation.cs
SerShrubert: Updated

src/Kerbalism/Science/ScienceSituation.cs
SerShrubert: It could be if the planet pack creator chooses to use 0. I don't know if that is valid normally. I've chosen to ignore invalid biome values and fallback to the existing body multiplier, rather than invalidating the whole situation and using the 1.0 fallback.

PR #1161, opened 2026-08-11
fixed lab experiment (aquaculture)

sherhan19:
+ the experiment row is moved to the end of the lab configuration list.
+ added a small mass and cost for placing experimental equipment in the laboratory.
+ added a functional requirement for the "Aquaculture" part from the SSPX mod.
+ added a small resource consumption by fish.


Aebestach:
Looks like the whole file was rewritten because line endings changed from LF to CRLF. Please convert back to LF so the real aquaculture experiment changes are easier to review.

sherhan19:
done

GameData/KerbalismConfig/Support/SSPX_Science.cfg
```diff
@@ -220,15 +226,15 @@
...
+@PART[*]:HAS[@MODULE[Configure]:HAS[#configureId[LaboratoryExperiments]]]:NEEDS[FeatureScience,StationPartsExpansionRedux]:FINAL
 {
 	@MODULE[Configure]:HAS[#configureId[LaboratoryExperiments]]
 	{
 		@SETUP:HAS[#name[Fish?Species?Observation]]
 		{
-			%tech = #$../../@KERBALISM_EXPERIMENT_VALUES/SSPX/FishObservation/UnlockTech$
-			%mass = #$../../@KERBALISM_EXPERIMENT_VALUES/SSPX/FishObservation/SetupMass$
-			%cost = #$../../@KERBALISM_EXPERIMENT_VALUES/SSPX/FishObservation/SetupCost$
+			%tech = advScienceTech
```
Aebestach: The original logic pulled from `KERBALISM_EXPERIMENT_VALUES/.../UnlockTech`, and the CTT patch in the same file would change this to `longTermHabitation`. After the PR hardcoded it to `advScienceTech`, CTT users would incorrectly unlock Fish Species Observation as if using the stock tech tree.

GameData/KerbalismConfig/Support/SSPX_Science.cfg
```diff
@@ -31,11 +31,11 @@
 			size = 21832
 			value = 24								
 			duration = 3888000							// 180 days
-			requirements = 								//Part:sspx-aquaculture-375-1 - broken requirement.
+			requirements = Part:sspx-aquaculture-375-1				//Part:sspx-aquaculture-375-1 - broken requirement.
 			CrewRequirement = Scientist@4
-			ResourceRates =								// quarter of resource consumption of a kerbal. Fish need oxygen, water, food to stay alive, i guess.
-			SetupMass = 0								// because the experiment requires the aquaculture part on the vessel, there's no added cost or mass to the lab setup.
-			SetupCost = 0
+			ResourceRates =	Water@0.000156,Oxygen@0.000156,Food@0.000156		// quarter of resource consumption of a kerbal. Fish need oxygen, water, food to stay alive, i guess.
```
Aebestach: Could you confirm whether the rates should be consistent here? This doesn't seem to make sense.

Aebestach (changes requested):

GameData/KerbalismConfig/Support/SSPX_Science.cfg
SerShrubert: Should these rates be on the aquaculture part itself? Fish don't stop consuming resources because you stop studying them.

sherhan19: Yes, that looks logical. But we have to choose: do we breed fish for food (greenhouse) or study them in the lab? Can Kerbals study them and eat them at the same time? :)

Aebestach (approved):

GameData/KerbalismConfig/Support/SSPX_Science.cfg
```diff
@@ -199,36 +199,45 @@
...
 }
 
 // =========================================================================
-// Patching the SETUP.
+// Patching the SETUP (Mass, Cost, Tech)
+// Uses global path #$@...$ so FINAL can still read KERBALISM_EXPERIMENT_VALUES
+// (including the CTT UnlockTech override above).
 // =========================================================================
-@PART[*]:HAS[@MODULE[Configure]:HAS[#configureId[LaboratoryExperiments]]]:NEEDS[FeatureScience,StationPartsExpansionRedux]:FOR[KerbalismDefault]
+@PART[*]:HAS[@MODULE[Configure]:HAS[#configureId[LaboratoryExperiments]]]:NEEDS[FeatureScience,StationPartsExpansionRedux]:FINAL
```
SerShrubert: Avoid FINAL on parts that don't belong to Kerbalism. This can be a :AFTER[zzzKerbalismDefault].
There will be a @PART[*]:HAS[@MODULE[Configure]:HAS[#configureId[LaboratoryExperiments]]]:NEEDS[FeatureScience]:FOR[zzzKerbalismDefault] in Patches-Experiments.cfg, but I will be proposing to drop the global lab multipliers.

GameData/KerbalismConfig/Support/SSPX_Science.cfg
SerShrubert: kerbalism_CHILLED requires the greenhouse but doesn't have any consumption of its own. I think that would make sense for the aquaculture experiment too.

Aebestach:
Combine `Fish Species Observation` two changes into one.

GameData/KerbalismConfig/Support/SSPX_Science.cfg
sherhan19: It would make more sense to rewrite the resource consumption for the aquaculture module like this: Currently, it consumes water, carbon dioxide, ammonia, and electricity, while producing food and oxygen. Instead, it should consume water, oxygen, food, and electricity, and produce food, carbon dioxide, and ammonia (or wastewater). But that's only if anyone actually needs it. Since I've only ever seen one module like this, it's fine to just leave it as is.

PR #1182, opened 2026-08-18
Fix excessive EC drain when transmitting experiments

SerShrubert:
Experiment.FixedUpdate is one cycle behind Kerbalism.FixedUpdate. 
This can cause excessive EC drain when transitioning from high bandwidth comms to very low bandwidth comms.
Experiment puts data on transmit buffer assuming a large capacity. When comms drop to very low bandwidth on the next cycle the existing data on the transmit buffer now has a very large EC cost to transmit. 4 million EC in one observed case.

Move Experiment.FixedUpdate to Kerbalism.FixedUpdate to match background experiment processing.
Also update profiler strings to aid in profiling with the Unity analyzer.

Iterating through the cached experiments took 17 us on a vessel with 36 total Experiment modules and 17 enabled Experiment modules. Of those 7 were active and waiting and took 197us for all of the fixedUpdate calls. Room for improvement here.

### DIFF
--- a/src/Kerbalism/System/Kerbalism.cs
+++ b/src/Kerbalism/System/Kerbalism.cs
@@ -318,7 +318,7 @@ namespace KERBALISM
 				// if loaded
 				if (v.loaded)
 				{
-					Profiler.BeginSample("Loaded.VesselDataEval");
+					Profiler.BeginSample("Kerbalism.Loaded.VesselDataEval");
 					// update the vessel info
 					vd.Evaluate(false, elapsed_s);
 					Profiler.EndSample();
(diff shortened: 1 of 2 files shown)


src/Kerbalism/System/Kerbalism.cs
```diff
@@ -318,42 +318,49 @@ void FixedUpdate()
...
 					Profiler.EndSample();
 
-					Profiler.BeginSample("Loaded.Comms");
+					Profiler.BeginSample("Kerbalism.Loaded.Comms");
 					CommsMessages.Update(v, vd, elapsed_s);
 					Profiler.EndSample();
 
-					Profiler.BeginSample("Loaded.Science");
+					Profiler.BeginSample("Kerbalism.Loaded.Experiment");
+					foreach(Experiment e in PartModuleCache.GetModules<Experiment>(v))
```
Aebestach, line 342: Should we skip Experiment modules that have already been disabled?

PR #1188, opened 2026-08-19
Fix BDB Skylab compatibility; rebalance lab experiment multipliers (all labs)

SerShrubert:
Fixed some issues mostly related to Skylab for BDB

- The crew reset requirement on the Apollo Telescope Mount was not working, removed it
- Reduce mass on parts that get extra supply. This mostly applies to Skylab which launched with a large water supply. 
- Skylab is large enough that both laboratory and crew experiments get added. The existing lab multiplier applied to all experiments on the part, including the crew experiments.
- I dug into the commit history and found that the values in StockGroups-Lab.cfg were intended to be the final [values.](https://github.com/Kerbalism/Kerbalism/commit/1f6316ee4f4e81f4aa9fb3cb3aee6f0f27ef87f9) Thus the lab multipliers are applied twice. To avoid breaking existing saves by doubling the experiment duration, I left the duration multipliers as is but reduced the EC multiplier from 5x to 2x to bring them down to more reasonable values. Mystery goo and materials study are short enough that they didn't need the duration reduction in the first place, but they kept their sample capacity and reset ability.

### DIFF
--- a/GameData/KerbalismConfig/System/ScienceRework/Patches-Experiments.cfg
+++ b/GameData/KerbalismConfig/System/ScienceRework/Patches-Experiments.cfg
@@ -397,14 +397,9 @@
 
 @PART[*]:HAS[@MODULE[Configure]:HAS[#configureId[LaboratoryExperiments]]]:NEEDS[FeatureScience]:FOR[zzzKerbalismDefault]
 {
-	// Following #775, this doesn't make any sense to me since lab experiments are only on labs, 
-	// why mass-patch their defined stats here ? Leaving as is because I'm lazy and I don't care.
-	@MODULE[Experiment],*
+	// Remove crew reset requirements for laboratory experiments
+	@MODULE[Experiment]:HAS[#experiment_ID[mobileMaterialsLab,mysteryGoo]]
 	{
-		@data_rate *= #$@KERBALISM_GROUP_SETTINGS/LAB_EXPERIMENTS/LabDataRateMultiplier$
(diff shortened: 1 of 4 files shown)


GameData/KerbalismConfig/System/ScienceRework/Patches-Experiments.cfg
```diff
@@ -397,14 +397,9 @@
...
-	// Following #775, this doesn't make any sense to me since lab experiments are only on labs, 
-	// why mass-patch their defined stats here ? Leaving as is because I'm lazy and I don't care.
-	@MODULE[Experiment],*
-	{
-		@data_rate *= #$@KERBALISM_GROUP_SETTINGS/LAB_EXPERIMENTS/LabDataRateMultiplier$
-		@ec_rate *= #$@KERBALISM_GROUP_SETTINGS/LAB_EXPERIMENTS/LabECMultiplier$
-		// removed, see issue #775, don't override lab experiments crew reqs
-		// %crew_operate = Scientist  
+	// Remove crew reset requirements for laboratory experiments
+	@MODULE[Experiment]:HAS[#experiment_ID[mobileMaterialsLab,mysteryGoo]]
```
Aebestach: `experiment_id`
I remember, `HAS` is case-sensitive.  The others part of this file uses `experiment_id`.

Aebestach (changes requested):

GameData/KerbalismConfig/System/ScienceRework/Patches-Experiments.cfg
SerShrubert: Fixed

Aebestach (approved):

PR #1197, opened 2026-08-24
Use KSPCF Fast/ReadOnly APIs for loaded part module lookup

Aebestach:
## Change Description
- Add KSPCF as a compile-time reference (already required at runtime).
- Use Fast / HasModuleImplementingFast / ReadOnly for loaded parts instead of allocating `FindModulesImplementing` lists.
- Name-filtered walks stay on `part.Modules`.


src/Kerbalism/Automation/Devices/Habitat.cs
```diff
@@ -6,7 +6,7 @@ public sealed class HabitatDevice : LoadedDevice<Habitat>
 
 		public HabitatDevice(Habitat module) : base(module)
 		{
-			hasGravityRing = module.part.FindModuleImplementing<GravityRing>() != null;
+			hasGravityRing = module.part.FindModuleImplementingFast<GravityRing>() != null;
```
siimav: HasModuleImplementingFast

PR #1206, opened 2026-09-06
Update Kill method to work better with Bureaucracy...

R-T-B:
...By flagging all kerbals dead in the roster if they well, die.

As a side note, should we be touching rep at all?  This isn't a tycoon mod.  Bureaucracy users will probably want to set this to 0.

### DIFF
--- a/src/Kerbalism/System/Kerbalism.cs
+++ b/src/Kerbalism/System/Kerbalism.cs
@@ -1021,7 +1021,7 @@ namespace KERBALISM
 			return false;
 		}
 
-		// kill a kerbal
+		//  a kerbal
 		// note: you can't kill a kerbal while iterating over vessel crew list, do it outside the loop
 		public static void Kill(Vessel v, ProtoCrewMember c)
 		{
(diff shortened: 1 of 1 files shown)


src/Kerbalism/System/Kerbalism.cs
```diff
@@ -1021,7 +1021,7 @@ public static bool IsRescueMission(Vessel v)
 			return false;
 		}
 
-		// kill a kerbal
+		//  a kerbal
```
siimav: Not sure why you redacted the *action*

src/Kerbalism/System/Kerbalism.cs
```diff
@@ -1039,7 +1039,7 @@ public static void Kill(Vessel v, ProtoCrewMember c)
 						{ part = p; break; }
 					}
 
-					KerbalEVA kerbalEVA = part.FindModuleImplementingFast<KerbalEVA>();
+					KerbalEVA kerbalEVA = part.FindModuleImplementing<KerbalEVA>();
```
siimav: That change is kinda unnecessary

siimav (changes requested):
Also a bunch of space -> tab conversions in this PR

src/Kerbalism/System/Kerbalism.cs
R-T-B: me neither, lol, I'll correct that when fixing the below.

src/Kerbalism/System/Kerbalism.cs
R-T-B: Uh yeah, caught me working with an old source copy and paste, fixing.

siimav (approved):

